PR #8, opened 2020-02-09
Fix urls in README.md

asashour:
Fix the link to `label-studio`, and update the link to `converter`


deppp:
Thanks for the PR @asashour !

PR #10, opened 2020-02-11
Fixes build failure in the default npm run start flow

ncryer:
When building the project as per instructions in README.md one currently encounters two errors: 

1. 
 ```bash
> bash scripts/copy.sh
cp: target 'src/examples/' is not a directory 
``` 
Once fixed, one gets: 
2. 
```
Failed to compile

./src/examples/image_bbox/index.js
Module not found: Can't resolve './completions/1.json' in 'path/to/label-studio-frontend/src/examples/image_bbox' 
``` 
 
This PR fixes both issues by tweaking `scripts/copy.sh` and by adding the missing completion data

### DIFF
--- a/scripts/copy.sh
+++ b/scripts/copy.sh
@@ -1,2 +1,2 @@
-mkdir src/examples
+mkdir -p src/examples
 cp -r examples/* src/examples/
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: scripts/copy.sh [hunk 1]
```diff
@@ -1 +1,2 @@
+mkdir src/examples
```
deppp: I think that one should be 

```bash
mkdir -p src/examples
```

deppp:
Thanks for your PR @ncryer, merging..

PR #11, opened 2020-02-12
Ellipse shape for selection area & universal command for build on win/unix/os x

lrlunin:
This patch adds functionality to label an image with the ellipse-shaped area. Also I have fixed the command for build from `CI=true` to `set CI=true`, cause the first command doesn't work on windows machines.


deppp:
@lrlunin First of all, absolutely great work! Here are a couple of minor things I've noticed that needs attention:

- mathjs is not in the package.json
- saved completions are not loaded for some reason
- the Node name `Ellipse 97.00 x 128.00, θ = 52.31793070243489, center = (474.00,580.00)` I'd format the theta to two points after the dot

All in all, it looks just a step away from being merged! Thanks for your contribution!

lrlunin:
@deppp Sorry for missed bugs. Now I have completely tested it with different examples. It should work fine now. As I noticed mathjs is in package.json. Indeed this library will be not used and can be deleted from packaging and source files. Should I do it?

PR #12, opened 2020-02-20
Added an ability to pan image when zoomed

olegtarasov:
Hi!

We've got this project where we need to detect lots of small objects in a photo. We discovered Label Studio to be a great tool, with one downside: you can't pan the image while it's zoomed. We use zoom all the time and constantly zooming in and out is really unproductive. So I've added an ability to pan the image along with its polygons. When image is zoomed, it can be panned with Shift + Left mouse button, or with middle mouse button.

### DIFF
--- a/src/components/ImageView/ImageView.js
+++ b/src/components/ImageView/ImageView.js
@@ -66,7 +66,18 @@ export default observer(
        */
       item.freezeHistory();
 
-      return item.onMouseMove(e);
+      const stage = item.stageRef;
+      const scale = stage.scaleX();
+
+      if (e.evt && (e.evt.buttons == 4 || (e.evt.buttons == 1 && e.evt.shiftKey)) && scale > 1) {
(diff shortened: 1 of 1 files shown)


deppp:
@olegtarasov Perfect! I'll check this during the weekend and will merge. Thanks for the contribution!

aiyou9:
Thank you for updating this function, which is very helpful to me. But a new function requirement was needed for my image project. Because there are hundreds of small targets in a single image, would you have a plant to provide a function button to locking the current select tag in the next version？best wish ！

deppp:
> Thank you for updating this function, which is very helpful to me. But a new function requirement was needed for my image project. Because there are hundreds of small targets in a single image, would you have a plant to provide a function button to locking the current select tag in the next version？best wish ！

@aiyou9 Hey, yeah, I've thought about `cue` functionality, maybe next release

PR #27, opened 2020-03-12
Fix lint warnings

jverre:
I noticed some lint warnings that caused that our CI process would fail.

I fixed the majority of them (mostly unused imports and equality signs), for the ones I wasn't sure about, I used `// eslint-disable-line`

I didn't run any tests apart from the test suite and making sure all the examples were still running.


deppp:
@jverre Excellent! This is something I was planning to do for a while, so thanks for taking care of it!

PR #44, opened 2020-03-28
feat: add rgbd support. Press 'tab' to switch between rgb and depth.

chenynCV:
Add rgbd support. So you can label based on rgb and depth input. Please have a look.

### DIFF
--- a/examples/image_rgbd/START.md
+++ b/examples/image_rgbd/START.md
@@ -1,0 +1,32 @@
+
+# Image object detection 
+
+![Image Object Detection](/images/screenshots/image_bbox.png "Image Object Detection")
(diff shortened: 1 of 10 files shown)


deppp:
@chenynCV Hey, thanks for the contribution, however, I was not able to make it work. When I click on the button it shows white container. Can you please attach a screenshot of what it looks like on your side?

chenynCV:
@deppp Awsome work first. 
 I add an example to show how to make it work.  You should pass `value2` to switch between rgb and depth image. And the two images size should be the same.
But when you press `tab` you will see black probably... Because the two images' size are not the same.

PR #50, opened 2020-04-02
fix zoom icon

chenynCV:
fix zoom icon by adding `< />`


hlomzik:
fixed already in https://github.com/HumanSignal/label-studio-frontend/pull/52

PR #62, opened 2020-04-10
Feature/rotate image

hlomzik:
- [x] rotate and zoom
- [x] save and restore rotation after switching to other completion/prediction
- [x] add cw rotation button
- [x] fix "view all" mode
- [x] serialisation should work correctly ~~in one of these ways:~~
  a. store regions as they were not rotated
  ~~or b. serialise and restore image rotation~~


rishabmps:
Hi, Will this be merged in the coming release?

hlomzik:
Not in the nearest releases I afraid. We have to make some decisions about serialising data and using it in the platform.
I can merge master into this branch, so you can use it directly.

PR #67, opened 2020-05-05
Feature/settings continue labeling

hlomzik:
Add new settings "Keep label selected after creating a region".
**When selected**: labels stay selected after new region created.
**Check existed functionality**: Labels are not selected after deserialisation.

Also improved some related code. DRY a bit.

### DIFF
--- a/src/components/ImageView/LabelOnRegion.js
+++ b/src/components/ImageView/LabelOnRegion.js
@@ -103,8 +103,6 @@ const LabelOnPolygon = observer(({ item }) => {
           shadowBlur={0}
         />
       )}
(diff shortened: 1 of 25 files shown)


src/components/ImageView/LabelOnRegion.js
```diff
@@ -103,8 +103,6 @@ const LabelOnPolygon = observer(({ item }) => {
           shadowBlur={0}
         />
       )}
-      {/* TODO this is a problem here, it takes states[0] but nothing
-       * guarantees that this state is a lable one */}
```
deppp, line 107: Good point. basically, the controls tags need to implement the interface and here we would just look up the state that provide that

PR #68, opened 2020-05-06
Show speed in a corresponding select for Audio

hlomzik:


src/env/development.js
```diff
@@ -41,7 +41,7 @@ import { Pairwise } from "../examples/pairwise"; // eslint-disable-line no-unuse
  */
 // import { AllTypes } from "../examples/all_types"; // eslint-disable-line no-unused-vars
 
-const data = ImageBbox;
+const data = AudioRegions;
```
deppp, line 44: This line is constantly gets changed and committed. We need to think of how that can be configured and edited locally.

PR #82, opened 2020-05-18
Builds are moved: git => GitHub Actions artifacts

hlomzik:
Now builds are generated automatically on push and placed as artifacts
on corresponding GitHub actions. Unless you add [skip ci] prefix.

Merge after #78 merged


hlomzik:
Should be recreated from scratch because files were changed

PR #83, opened 2020-05-19
Add maxUsages for Label/Labels

hlomzik:
Controls how many times Label can be used.

- [x] Text
- [x] Image
- [x] others


hlomzik:
> @hlomzik Andrew, do we also need to add here some code to support that functionality in HyperText & Paragraphs?

Added it to AudioPlus and HyperText. Paragraphs in the other PR, so skipped it.

PR #94, opened 2020-05-28
Hotfix/atob unicode

deppp:


hlomzik:
base64unicode encoding itself is in the master branch already. Other stuff in other branches.

PR #112, opened 2020-06-12
Don't allow shapes to move/scale beyond the Image borders

hlomzik:
It's fine, but there is a huge problem with rotation — we have to calculate bounding rect.
And do we really need it? Shape can be used to label some object near or on the border.


hlomzik:
Implemented with a better approach considering rotate. Partially already in master, partially inside branch, we'll be merged in soon.

PR #125, opened 2020-07-08
Don't unselectAll during initialization

hlomzik:


hlomzik:
Useless now — tmp/deserialization improves all these processes.

PR #134, opened 2020-08-21
added reverse zoom option (scroll up == zoom in)

wdvr:
Added a parameter to configure zoom direction of the scroll wheel.

Works for me on the examples.


wdvr:
See https://github.com/heartexlabs/label-studio-frontend/issues/135 for explanation of the issue

This is a proposal to keep the current behavior as is and not change current implementations. Nicer/easier solution would be to simply reverse the sign check of `e.evt.deltaY`

PR #137, opened 2020-08-24
New tag: RichText

nicholasrq:
Meet `RichText`, a *drop-in replacement* for `Text` and `HyperText`.

### Features
* Support both text and HTML with back compatibility to `HyperText` and `Text` output.
* Loading value from URL
* Experimental granularity options: `sentence` and `paragraph` _(only Chrome and WebKit for now)_
* Create regions programmatically. No more `needsUpdate` mess.
* Efficient and nice :)

### Optimizations

* Avoid heavy loops when traversing the DOM. Instead use [TreeWalker API][TreeWalker] that is designed specifically for that task.
* Avoid repetitive event bounding. `RichText` delegates all the events to its root node instead of binding events to every region directly. It means that event handlers are there from the beginning and we don't touch them ever.
* Use dynamic stylesheets instead of inline styles. This approach allows `RichText` to change styles of regions without DOM lookup. We only add/remove classes on specific nodes when needed, in other cases it's a matter of changing global [CSS Variables][CSSVars].
* Overall reduction of DOM mutations.

#### The approach
`RichText` uses previous approach to highlight text by utilizing `Range`, but does it in a different way. When possible, it avoids nested loops (and loops in general) to build a highlighted region.

#### Efficiency
`RichText` tries to reduce number of DOM manipulation and tree traversal, instead it uses `TreeWalker` – native DOM utility to walk through huge amount of nodes efficiently. It's a generator-based flat representation of DOM with native filtering. It allows to select nodes very precisely avoiding unnecessary operations.

#### Styling
For styling `RichText` creates dynamic stylesheets and manipulates css-variables instead of writing styles inline. It helps to avoid unnecessary DOM manipulations as well.

#### Highlights and intersecting regions
`RichText` is always trying to keep the DOM consistent. To achieve that it uses bottom-up selecting approach. It means that smaller selection will always be inside a larger selection. This approach also allowed to remove excess event handlers.

By utilizing dynamic stylesheets `RichText` can highlight nodes by adding/removing classes to regions. Along with that regions and content around don't jump anymore when highlighted.

ezgif-6-7920dd951f3a

### Programmatic regions creation

Now you can do this:

```javascript
Htx.completionStore.selected.names.get('text').createRegion({
    "id": "Gm23vpxJQE",
    "from_name": "ner",
    "to_name": "text",
    "source": "$text",
    "type": "labels",
    "parent_id": null,
    "value": {
        "start": 291,
        "end": 308,
        "text": "will be late this",
        "labels": [
            "Date"
        ]
    }
})
```

and this:

```javascript
Htx.completionStore.selected.names.get('text').createRegion({
    "from_name": "ner",
    "id": "beFiQTT1cU",
    "source": "$text",
    "to_name": "text",
    "type": "htmllabels",
    "value": {
        "endOffset": 97,
        "startOffset": 86,
        "start": "/text()[2]",
        "end": "/text()[2]",
        "htmllabels": [
            "Date"
        ],
        "text": "February 22"
    }
})
```

#### Cleanup
After region removal `RichText` will restore original DOM structure and remove unused styles to avoid glitches.

#### Try it

To try it out you can use examples in `development.js`, there are `RichTextPlain` and `RichTextHtml` to showcase work with text and html accordingly.

[TreeWalker]: https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker
[CSSVars]: https://developer.mozilla.org/en-US/docs/Web/CSS/var


hlomzik (commented):
Small note: remove logs

src/tags/object/Text.js
```diff
@@ -85,6 +85,10 @@ const Model = types
       self._ref = ref;
     },
 
+    needsUpdate() {
+      self._update = self._update + 1;
+    },
```
hlomzik: thanks! I was about to ask you about this :)

src/tags/object/RichText/model.js
```diff
@@ -0,0 +1,272 @@
...
+
+          Infomodal.error(message.map(t => <p>{t}</p>));
+          self.setRemoteValue("");
+          return;
+        }
+
+        const response = await fetch(url);
+        const { ok, status, statusText } = response;
+
+        if (!ok) throw new Error(`${status} ${statusText}`);
```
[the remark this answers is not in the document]
nicholasrq: oh, you're right, thanks

nicholasrq:
@hlomzik RichText is already registered and, as it's mentioned in the description to this PR, it's drop-in replacement for both HyperText and Text :) so yeah, it's ready to use. you can actually try it if you want, I added few examples

speaking of xpath, now I understand what "breaks DOM in weird ways" means :)

I was thinking, that we can generate xpath during the serialization stage because it's only needed in the serialized data and for regions restorations on the launch of LS. i'm still not sure if `Region` is dynamic in terms of DOM mutations and does it update its properties, I'll check it

if it's not, there's a way in utilizing `MutationObserver`. it allows to watch only particular changes on a selectd node and doesn't cost too much in terms of performance (especially when we don't need anything but subtree). basically, we can watch mutations and adjust Regions on the fly or generate regions form `span` positions directly (every region has and ID that tells, what regions it depends on). this will allow reduce the computations overall and will allow us to watch only intersecting regions.

and i believe that the approach with MutationObserver could be easily combined with xpath generation on the serialization stage to avoid constat changes during labeling. as we don't need xpath for labeling, I see no reason to calculate it there

and if none of this work, we can also experiment with the approach you mentioned. it might cause extra performance drops, but should be the most bulletproof.

src/tags/object/Text.js
nicholasrq: yeah, i noticed this one when was removing excess logs :) it was supposed to be removed in `RichText` but i missed the file, probably

src/mixins/HighlightMixin.js
```diff
@@ -0,0 +1,283 @@
...
+
+  /**
+   * Remove stylesheet
+   */
+  const remove = () => {
+    styleTag.remove();
+  };
+
+  return {
+    className: className.split(".").filter(el => !!el),
```
hlomzik, line 277: but `className` is set as constant, so you can just `substr(1)`

nicholasrq: yeah, you're right. there were two classes at the beginning, but I dropped one and this is just an artifact. thanks!

nicholasrq:
@hlomzik oh, perhaps I misunderstood your point about replacing `Text` and `HyperText`. did you mean that we can override `Text` and `HyperText`, so they will actually call `RichText` instead of their own implementation? if so, am I correct that it could be achieved just by registering `RichText` multiple times with different names?

nicholasrq:
@hlomzik hmm.. I just realized, that if we move xpath building to serialization stage, I can just use existing spans to calculate proper xpaths for the whole thing without even touching observers and other complex things. we already have everything we need at this point, even if native `Range` is broken in some way. I just need to take first and last spans of every range and either calculate offsets relative to root node for plain text, or build xpaths for html

hlomzik:
> @hlomzik oh, perhaps I misunderstood your point about replacing `Text` and `HyperText`. did you mean that we can override `Text` and `HyperText`, so they will actually call `RichText` instead of their own implementation? if so, am I correct that it could be achieved just by registering `RichText` multiple times with different names?

Yes, that what I mean! :)
And I've tried — it works. I think that we don't need one more tag with the same features, so better to leave the tag as it is, and just register it by different names. The only caveat — valueType should be assigned differently for these tags.

nicholasrq:
@hlomzik got ya! not a big deal, I'll fix valueType to make it match with other tags :) originally i renamed this attribute to clearly align its name with what it actually does

hlomzik:
Yes, I liked this name, it's more relevant!
But also I mean that html/text should be detected automatically somehow.
Actually during building a tree the `type` is assigned to lowercased name of the tag, so your `RichText` will have `type=(text|hypertext)` and you can use this to detect `isText`.

nicholasrq:
@hlomzik oh, that's clever! `RichText` does detect html/text using input data, but it's done empirically and depends on the input, so this approach doesn't work when we don't have input (nothing's labeled), or when we're adding new labels. leveraging the `name` would be the most robust solution here

nicholasrq:
@hlomzik  ok, I replaced `HyperText`, and `Text` with `RichText` implementation, and removed all artifacts from of `RichText`. `RichText` cannot be used as an independent tag now, but `Text`, and `HyperText` now use it as a backend. 

speaking of serialization, I want to try different approaches to understand which one works better. for the sake of backward compatibility I will leave current implementation as is, and will introduce an alternative later as a separate mechanism. it will be under a hidden attribute or something (you will need to set a specific attribute to enable new serialization). it will allow us to safely test it before rolling out. once we're sure that it works, we can just remove the attribute and make new behavior a default one

nicholasrq:
Closed in regard of #143

PR #143, opened 2020-10-01
New tag: RichText

nicholasrq:
Meet `RichText`, a *drop-in replacement* for `Text` and `HyperText`.

### Features
* Support both text and HTML with back compatibility to `HyperText` and `Text` output.
* Loading value from URL
* Experimental granularity options: `sentence` and `paragraph` _(only Chrome and WebKit for now)_
* Create regions programmatically. No more `needsUpdate` mess.
* Efficient and nice :)

### Optimizations

* Avoid heavy loops when traversing the DOM. Instead use [TreeWalker API][TreeWalker] that is designed specifically for that task.
* Avoid repetitive event bounding. `RichText` delegates all the events to its root node instead of binding events to every region directly. It means that event handlers are there from the beginning and we don't touch them ever.
* Use dynamic stylesheets instead of inline styles. This approach allows `RichText` to change styles of regions without DOM lookup. We only add/remove classes on specific nodes when needed, in other cases it's a matter of changing global [CSS Variables][CSSVars].
* Overall reduction of DOM mutations.

#### The approach
`RichText` uses previous approach to highlight text by utilizing `Range`, but does it in a different way. When possible, it avoids nested loops (and loops in general) to build a highlighted region.

#### Efficiency
`RichText` tries to reduce number of DOM manipulation and tree traversal, instead it uses `TreeWalker` – native DOM utility to walk through huge amount of nodes efficiently. It's a generator-based flat representation of DOM with native filtering. It allows to select nodes very precisely avoiding unnecessary operations.

#### Styling
For styling `RichText` creates dynamic stylesheets and manipulates css-variables instead of writing styles inline. It helps to avoid unnecessary DOM manipulations as well.

#### Highlights and intersecting regions
`RichText` is always trying to keep the DOM consistent. To achieve that it uses bottom-up selecting approach. It means that smaller selection will always be inside a larger selection. This approach also allowed to remove excess event handlers.

By utilizing dynamic stylesheets `RichText` can highlight nodes by adding/removing classes to regions. Along with that regions and content around don't jump anymore when highlighted.

ezgif-6-7920dd951f3a

### Programmatic regions creation

Now you can do this:

```javascript
Htx.completionStore.selected.names.get('text').createRegion({
    "id": "Gm23vpxJQE",
    "from_name": "ner",
    "to_name": "text",
    "source": "$text",
    "type": "labels",
    "parent_id": null,
    "value": {
        "start": 291,
        "end": 308,
        "text": "will be late this",
        "labels": [
            "Date"
        ]
    }
})
```

and this:

```javascript
Htx.completionStore.selected.names.get('text').createRegion({
    "from_name": "ner",
    "id": "beFiQTT1cU",
    "source": "$text",
    "to_name": "text",
    "type": "htmllabels",
    "value": {
        "endOffset": 97,
        "startOffset": 86,
        "start": "/text()[2]",
        "end": "/text()[2]",
        "htmllabels": [
            "Date"
        ],
        "text": "February 22"
    }
})
```

#### Cleanup
After region removal `RichText` will restore original DOM structure and remove unused styles to avoid glitches.

#### Try it

To try it out you can use examples in `development.js`, there are `RichTextPlain` and `RichTextHtml` to showcase work with text and html accordingly.

[TreeWalker]: https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker
[CSSVars]: https://developer.mozilla.org/en-US/docs/Web/CSS/var


src/mixins/HighlightMixin.js
```diff
@@ -0,0 +1,290 @@
...
+     */
+    setCursor(cursor) {
+      self._stylesheet.setCursor(cursor);
+    },
+
+    /**
+     * Draw region outline
+     * @param {boolean} val
+     */
+    setHighlight(val) {
```
hlomzik, line 106: setHighlighted?

src/stores/AnnotationStore.js
```diff
@@ -615,6 +620,19 @@ const Annotation = types
         self.list.addErrors([errorBuilder.generalError(e)]);
       }
     },
+
+    prepareValue(value) {
+      if (value.start && value.end && !value.startOffset && !value.endOffset) {
```
hlomzik, line 625: is it for back compatibility? could you please add comment for this?

src/tags/object/RichText/model.js
```diff
@@ -0,0 +1,247 @@
...
+
+        area.updateOffsets(soff, eoff);
+      }
+
+      area.applyHighlight();
+
+      return area;
+    },
+
+    fromStateJSON(obj, fromModel) {
```
hlomzik, line 203: not used so can be removed

src/tags/object/RichText/model.js
```diff
@@ -0,0 +1,247 @@
...
+
+    setRemoteValue(val) {
+      self.loaded = true;
+
+      if (self.encoding === "base64") val = atob(val);
+      if (self.encoding === "base64unicode") val = Utils.Checkers.atobUnicode(val);
+
+      self._value = val;
+
+      self._regionsCache.forEach(({ region, completion }) => {
```
hlomzik, line 139: I hope that this cache is redundant in the new engine.
it was used to store regions before the full load by url and create correct spans only after that.

hlomzik (approved):

PR #153, opened 2020-11-18
Fix clicking on poly borderline

niklub:


hlomzik:
Already done in `release/0.8.1`

PR #164, opened 2021-01-16
Show a few alternative versions of the image

hlomzik:
### Problem

Some images data can have different versions or layers (color channels, MRI levels, etc), so it would be useful to switch between them for better object detection.

### Solution

data: allow `$image` to be an array of images; in this case show gallery with these images and allow to switch between them. Config left the same as usual

Screen Shot 2021-01-16 at 4 32 56


makseq:
Build for LS 1.0 

[build b8c8dd7950dd637c25575cd510b2863996a1425e.zip](https://github.com/heartexlabs/label-studio-frontend/files/6383864/build.b8c8dd7950dd637c25575cd510b2863996a1425e.zip)

quantumdot:
The proposed feature would be a very valuable addition when working with multi-modal images.

@makseq Any instructions on how to use the build you linked above with an existing label-studio installation?

zakajd:
@quantumdot  Here is building instructions I used to make it work with LS 1.0
1. Create new Anaconda env (`conda create -n ls python=3.8)
3. Pull LS from GitHub `git clone https://github.com/heartexlabs/label-studio.git`
4. Replace css and js files with files from another build. (Attached as zip in makseq comment) `label-studio/label_studio/frontend/dist/lsf`
5. Install LS. `pip install .`
6. Run `python ./label_studio/manage.py collectstatic` without this it wouldn't work
7. Set env params: `export LABEL_STUDIO_LOCAL_FILES_SERVING_ENABLED=true` & `export LABEL_STUDIO_LOCAL_FILES_DOCUMENT_ROOT="/your/path"`
8. Launch server and verify that everything works normally. `label-studio start`

quantumdot:
Hi @zakajd,

Thanks for your reply, but I am having trouble to get this to work. 

1. I cloned down the current master version of label-studio (1.0.2)
2. Copy the `static/js/*` and `/static/css/*` files from the linked zip archive above to `label-studio/label_studio/frontend/dist/lsf`
3. Run `pip install .` and then run `python ./label_studio/manage.py collectstatic`

After I start the label-studio server, everything looks fine (e.x. project configuration, organization management), but then when I go to the labeling interface for a project, I see a mostly (except top status bar) blank screen. Checking the console, I get a 404 for `/dm/js/main.js?v=` and a rejected request (due to mismatching MIME) for `/dm/css/main.css?v=`. Sometimes I can trigger an error such as `Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.`

I tried with the earlier label-studio v1.0.0, but the process fails with a key error for `collectstatic` when running your step 5.

PR #174, opened 2021-03-10
Added parameters to set range of y-axis for timeseries

brainslush:
This feature adds parameters to set fixed values for the range of the y-axis of a time series. The parameters are called "upperBound" and "lowerBound". It is sufficient to set only one of the parameters. The unset bound will respect the value of the parameter "fixedScalce".
As a sideeffect when the lower bound is greater than the upper bound, the graph will be inverted. This can be a positive or negative effect. It remains open for discussion if this is a wanted effect.


brainslush:
Realized that there possibly are some minor things that needs improvement here.

PR #183, opened 2021-04-06
Fix `required` for nested `Choices`

hlomzik:


hlomzik:
Fixes https://github.com/heartexlabs/label-studio/issues/765

PR #226, opened 2021-07-16
Feature/image max height

Gondragos:


hlomzik:
Was continued and finished in https://github.com/HumanSignal/label-studio-frontend/pull/411

PR #236, opened 2021-08-13
Improve docs

hlomzik:


src/tags/object/Audio.js
```diff
@@ -12,7 +12,7 @@ import { ErrorMessage } from "../../components/ErrorMessage/ErrorMessage";
 import { AnnotationMixin } from "../../mixins/AnnotationMixin";
 
 /**
- * Audio tag plays a simple audio file.
+ * Audio tag plays a simple audio file
```
smoreface, line 15: does the period break something?

src/regions/AudioRegion.js
```diff
@@ -50,6 +50,26 @@ const Model = types
...
+     * {
+     *   "original_length": 18,
+     *   "value": {
+     *     "start": 3.1,
+     *     "end": 8.2,
+     *     "labels": ["Voice"]
+     *   }
+     * }
+     * @typedef {Object} AudioRegionResult
+     * @property {number} original_length length of the original audio (s)
```
smoreface, line 64:
```suggestion
     * @property {number} original_length length of the original audio
```

src/regions/AudioRegion.js
```diff
@@ -50,6 +50,26 @@ const Model = types
...
+     * @property {Object} value
+     * @property {number} value.start start time of the fragment (s)
```
smoreface, line 66:
```suggestion
     * @property {number} value.start start time of the fragment(s)
```

src/regions/AudioRegion.js
```diff
@@ -50,6 +50,26 @@ const Model = types
...
+     * @property {number} value.end finish time of the fragment (s)
```
smoreface, line 67:
```suggestion
     * @property {number} value.end finish time of the fragment(s)
```

src/tags/object/Audio.js
hlomzik: of course no :)
we can add it back for sure

src/regions/AudioRegion.js
```diff
@@ -61,10 +61,10 @@ const Model = types
      *   }
      * }
      * @typedef {Object} AudioRegionResult
-     * @property {number} original_length length of the original audio (s)
+     * @property {number} original_length length of the original audio
```
hlomzik, line 64: That's a format of the value :)
seconds in this case
I put it for further automatic data generation
I'll return it back, but if you have any idea, how to make it better, I'd be glad to hear!

hlomzik: I'll use `(seconds)`

smoreface: perfect, thank you! I didn't realize that's what it was lol

smoreface (approved):

PR #249, opened 2021-08-23
IFrame for HyperText to isolate document in it

hlomzik:
- [x] serialization
- [x] deserialization
- [ ] fix wrong old xpath (should work, but have to check)
- [x] work for html in data (`valueType="text"`)
- [x] handle hotkeys
- [x] enable only by some param (like `embed="true"`)
- [ ] add example
- [ ] add test
- [ ] relations
- [x] styles (no border, max width, responsive)


hlomzik:
Added parameter `inline="false"` to use direct insert of text/html.
In `Text` tag it's always `true`. In `HyperText` it's `false` by default, so html is embedded as `iframe`.

PR #257, opened 2021-08-25
DateTime tag

hlomzik:
`<DateTime name="full" toName="audio" only="date" required="true" min="2021-08-10" format="%d.%m.%Y"/>`
`only` can be "date" or "time" to use only part of the datetime.
Appearance depends on browser and locale (so it's good for annotators), but value stored in given format or ISO by default.


src/tags/control/DateTime.js
```diff
@@ -0,0 +1,235 @@
...
+
+        return self.annotation.results.find(r => r.from_name === self && r.area === area);
+      }
+      return self.annotation.results.find(r => r.from_name === self);
+    },
+  }))
+  .volatile(self => {
+    const defaultFormatters = {
+      formatDate: d3.timeFormat(FORMAT_DATE),
+      formatTime: d3.timeFormat(FORMAT_TIME),
```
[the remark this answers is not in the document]
hlomzik: we already use d3 formatting in TimeSeries and it's actually strftime formatting, commonly used in python/ml development, so I'd stick with it. That's the formatting of the final result, so users can have the same format here and in there processing software.
Date formatting in the interface is powered by `input type=date`, so it's just based on user's locale. That could be changed later.

src/tags/control/DateTime.js
```diff
@@ -0,0 +1,338 @@
...
+ *   <Text name="txt" value="$text" />
+ *   <DateTime name="datetime" toName="txt" only="date" />
+ * </View>
+ *
+ * @name DateTime
+ * @param {string} name              - Name of the element
+ * @param {string} toName            - Name of the element that you want to label
+ * @param {string} only              - Comma-separated list of parts to display (date, time, month, year)
+ *        date and month/year can't be used together. The date option takes precedence
+ * @param {string} format            - Input/output strftime format for datetime (internally it's always ISO);
```
smoreface, line 38: not sure if it matters to point out that "internally it's always ISO" unless that means that ISO is what gets exported. 
Does input/output here mean that it's the formatting for either the input OR the output or the input AND the output? if the latter, plz change `/` to `and`

smoreface: can't add a suggestion but change it to:

strftime-formatted datetime string

hlomzik:
Merged by https://github.com/heartexlabs/label-studio-frontend/pull/713

PR #258, opened 2021-08-26
Get annotationStore from direct parents, not root

hlomzik:
There are may be some different weird errors when store is detached, so looking for annotationStore from root is not possible. But it's easier and more correct to walk through all the parent.


hlomzik:
fixed in a better way in https://github.com/HumanSignal/label-studio-frontend/pull/928

PR #272, opened 2021-09-08
Fixes for brushes and polygons

Gondragos:
- add "bboxCoords" computed value to the image's shape models
- make labels on region components for brushes and polygons dependent on the "bboxCoords" field
- improve brush regions performance
- fix erasing labels on brush region
- fix glitch while drawing brush
- fix page crashing on switching tasks with polygons


Gondragos:
Already merged in context of multi selection

PR #277, opened 2021-09-12
DEV-584: Fixes for view all

Gondragos:
- fix calculating ready states for image regions inside view all
- allow to display regions in richtext's iframes inside view all


Gondragos:
Still works wrong with some browsers

hlomzik:
was already merged inside https://github.com/heartexlabs/label-studio-frontend/pull/426 for DEV-1235

PR #290, opened 2021-09-17
Add support for all LTS node versions

twsl:
Content:
- Test against all current [LTS](https://nodejs.org/en/about/releases/) versions of node
- Don't fix yarn version
- Support for possibly test on multiple OS in the future


twsl:
Remark: Tests in general doesn't seem to be stable/reproducible.
It took 2 re-runs without change to get the pipeline to pass.

twsl:
@niklub @makseq can someone please review it?

PR #301, opened 2021-09-18
Restructure ImageView component to use canvas only

### LINKED ISSUE #299: Move the image in ImageView component to canvas
I'd like to propose the move of the actual image from an image tag to the canvas, mainly because it would make my feature request https://github.com/heartexlabs/label-studio/issues/1425 much easier to implement using the existing [filters](https://konvajs.org/docs/filters/Emboss.html) in konva.

While this might come at a minor performance penalty, I do think the advantage of interactive filters to support labeling justifies this move.

Any help implementing this is appreciated as I am currently having problems calculating the size of the image.

twsl:
This PR is intended to fix #299.
This is going to be a quite big restructuring and work in progress.

This PR will hopefully include:
- Remove img tag
- Introduce img as canvas layer using the Image component from konva
- Update all events to support resizing etc.


twsl:
Idea: use existing img tag to grab image, then load to canvas. Keeping the img tag prevents annoying stage sizing problems.
Problems now seems to bee canvas and [CORS](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image)

twsl:
Alright, CORS problem has been solved. Yet the main issue remains the size of the image area as it is currently only defined by the maxWidth property of the img parent. These CSS properties won't work for the canvas.

twsl:
The ImageView component now includes an additional canvas layer, if filters are enabled. Filters are optional, therefore the default img tag stays untouched. Basic filters are wroking but not configurable yet.

PR #318, opened 2021-10-14
Feature/tag choices list

AlexanderSkvira:
Problem: 
 Unable to pass data-dependent options to Choices tag.

Solution:
  Added ChoicesList tag with options taken from JSON data


hlomzik:
Already solved by "Dynamic Choices": you can use `<Choices value="$options" />` with `options` in data with the same props as usual Choices.

PR #338, opened 2021-11-01
DEV-839 Add keeping region selection on time travel

Gondragos:
This should fix the first problem from https://github.com/heartexlabs/label-studio/issues/1242


Gondragos:
closed due to https://github.com/heartexlabs/label-studio-frontend/pull/356

PR #340, opened 2021-11-10
[fix] Add observing of image size for special cases

Gondragos:
This should fix the canvas resizing when the scrollbar appears.


src/components/ImageView/ImageView.js
```diff
@@ -18,6 +18,7 @@ import { Toolbar } from "../Toolbar/Toolbar";
 import { ImageViewProvider } from "./ImageViewContext";
 import { Hotkey } from "../../core/Hotkey";
 import { useObserver } from "mobx-react-lite";
+import ResizeObserver from "resize-observer-polyfill";
```
nicholasrq, line 21: Remove imoprt

nicholasrq (changes requested):

src/components/ImageView/ImageView.js
```diff
@@ -536,6 +537,8 @@ export default observer(
 
     componentDidMount() {
       window.addEventListener("resize", this.onResize);
+      this.resizeObserver = new ResizeObserver(this.onResize);
```
hlomzik, line 540: Can you make a fallback with empty methods for older browsers

src/components/ImageView/ImageView.js
Gondragos: Replaced with fallback

nicholasrq (approved):

nicholasrq (approved):

niklub (approved):

PR #351, opened 2021-11-15
DEV-1118: Fix rect transform when zoomed in

nicholasrq:


src/components/ImageTransformer/ImageTransformer.js
```diff
@@ -89,36 +89,70 @@ export default class TransformerComponent extends Component {
...
 
-    // it's harder to fix sizes for rotated box, so just block changes out of stage
+    // // it's harder to fix sizes for rotated box, so just block changes out of stage
     if (rotation || isRotated) {
       const { x, y, width, height } = newBox;
       const selfRect = { x: 0, y: 0, width, height };
 
       // bounding box, got by applying current shift and rotation to normalized box
       const clientRect = getBoundingBoxAfterChanges(selfRect, { x, y }, rotation);
-      const fixed = fixRectToFit(clientRect, stage.width(), stage.height());
```
hlomzik, line 109: remove import of `fixRectToFit`

src/components/ImageTransformer/ImageTransformer.js
```diff
@@ -89,36 +89,70 @@ export default class TransformerComponent extends Component {
...
+      height = stage.height - realY;
+    }
+
+    return { ...box, x, y, width, height };
+  }
+
+  getStageAbsoluteDimensions() {
+    const stage = this.transformer.getStage();
+    const [scaledStageWidth, scaledStageHeight] = [stage.width() * stage.scaleX(), stage.height() * stage.scaleY()];
+    const [stageX, stageY] = [stage.x(), stage.y()].map(Math.abs);
```
hlomzik, line 117: would be more correct to leave them as is and subtract them in `fitBBoxToScaledStage()` because this math is based on assumption that we'll never move stage to positive coordinates, but we can change this later.

hlomzik (approved):

src/components/ImageTransformer/ImageTransformer.js
nicholasrq: fair point

niklub (approved):

PR #352, opened 2021-11-15
[fix] keep labels selected after region removal

hlomzik:
if "Keep labels selected" is active in settings


nicholasrq (approved):

hlomzik:
Duplicated #358

PR #355, opened 2021-11-16
DEV-1154: Fix scaling on rotation

nicholasrq:


src/components/ImageView/ImageView.js
```diff
@@ -239,6 +239,8 @@ const SelectionLayer = observer(({ item, selectionArea }) => {
         supportsTransform={supportsTransform}
         supportsScale={supportsScale}
         selectedShapes={item.selectedRegions}
+        singleNodeMode={item.selectedRegions.length === 1}
```
hlomzik, line 242: I think it's easier to calculate that inside, but doesn't matter

hlomzik (approved):

PR #362, opened 2021-11-19
DEV-1094 When I go back to previous skipped task, I can't change it

chiganov:


src/components/TopBar/Controls.js
```diff
@@ -56,8 +56,14 @@ export const Controls = controlsInjector(observer(({ store, history, annotation
   } else if (annotation.skipped) {
     buttons.push(
       <Elem name="skipped-info" key="skipped">
-        <IconBan color="#d00" /> Annotation skipped
-      </Elem>,
+        <IconBan color="#d00" /> Was skipped
+      </Elem>);
+    buttons.push(
+      <ButtonTooltip key="cancel-skip" title="Cancel skip: []">
+        <Button aria-label="cancel-skip" disabled={disabled} look="primary" onClick={() => store.annotationStore.deleteAnnotation(annotation)}>
```
nicholasrq, line 63: if someone got back to a cancelled annotation and made some labeling there before cancelling the skip, they will lose their work. that's a potential flaw. let's maybe add a warning here as a temporary solution

nicholasrq (changes requested):

src/components/TopBar/Controls.js
chiganov: It would be easier to patch `skipped` field. But there will be annotation without result.

farioas (approved):
My approval cover changes in .github/ dir

PR #381, opened 2021-12-06
DEV-1175 DEV-1220 Fixes for HyperText labeling: scripts, drafts

hlomzik:
- don't create drafts on HyperText project loading (that was because of offsets/xpaths fixes)
- remove all the scripts and iframes (previously multiline tags were not processed)


hlomzik:
will be fixed by #432 and further PRs

PR #397, opened 2021-12-16
DEV-1155: Rotation behavior

Gondragos:


src/components/ImageTransformer/ImageTransformer.js
```diff
@@ -82,34 +67,21 @@ export default class TransformerComponent extends Component {
     const prevNodes = this.transformer.nodes();
     // do nothing if selected node is already attached
 
-    if (!this.needsUpdate && selectedNodes?.length === prevNodes?.length && !selectedNodes.find((node, idx) => node !== prevNodes[idx])) {
+    if (
+      selectedNodes?.length === prevNodes?.length
+      && !selectedNodes.find((node, idx) => node !== prevNodes[idx])
```
nicholasrq, line 72: it's better to move such expressions to a separate variable with an explicit name of what it's doing

nicholasrq (approved):

PR #403, opened 2021-12-21
Update docs for HyperText/Text via stub files

hlomzik:


src/tags/object/Text.js
```diff
@@ -0,0 +1,34 @@
+// stub file to keep docs for Text object tag
+
+/**
+ * The Text tag shows text that can be labeled. Use to display any type of text on the labeling interface.
+ * You can use `<Style>.htx-text{ white-space: pre-wrap; }</Style>` to preserve all spaces in the text, otherwise spaces are trimmed when displayed.
```
smoreface, line 5: is this true for Text or just HyperText?

src/tags/object/HyperText.js
```diff
@@ -0,0 +1,35 @@
...
+ * @example
+ * <HyperText name="p1" value="<p>Some explanations <em>with style</em></p>" />
+ * @name HyperText
+ * @regions HyperTextRegion
+ * @meta_title Hypertext Tags for Hypertext Markup (HTML)
+ * @meta_description Label Studio Hypertext Tags customize Label Studio for hypertext markup (HTML) for machine learning and data science projects.
+ * @param {string} name                                   Name of the element
+ * @param {string} value                                  Value of the element
+ * @param {url|text} [valueType=text]                     Whether the text is stored directly in uploaded data or needs to be loaded from a URL
+ * @param {boolean} [inline=false]                        Whether to embed html directly to LS or use iframe
```
smoreface, line 26:
```suggestion
 * @param {boolean} [inline=false]                        Whether to embed HTML directly in Label Studio or use an iframe
```

src/tags/object/Text.js
smoreface: This implies that text written will be transformed from:
`text like this` to `textlikethis` 
but instead I think you mean
`text like      this` is transformed to `text like this` 
or?

hlomzik: That's classic html space trimming, I thought that trim is always about leading/trailing spaces, so no explanations needed. Also it collapses multiple spaces into one, so maybe we need some other term.
For `HyperText` that's expected behaviour, so this is helpful for `Text` only I think.

smoreface: Ok that makes sense. I think this is fine!

smoreface (approved):

PR #405, opened 2021-12-22
DEV-1175 Fixes for Drafts and undo/redo

hlomzik:
- HyperText starts already in draft state
- draft with relations broken on page reload
- undo/redo of audio region doesn't move or delete it
- after loading annotation created from prediction, moving some region and undo that you'll end up with empty annotation


src/stores/AnnotationStore.js
```diff
@@ -840,7 +840,7 @@ const Annotation = types
         });
       }
 
-      self.objects.forEach(obj => obj.needsUpdate?.());
+      self.updateObjects();
```
Gondragos, line 843: `updateObjects` contains `unselectAll`, so I'm afraid that this change may break the user experience or even the functional of suggestings in case of working with the ml-backend.

Gondragos (approved):

hlomzik:
fixed by #432 and #440

PR #411, opened 2021-12-27
DEV-1367: Image size problems

Gondragos:
- fix the calculation of the stage size on the window resizing at a given image width
- add frame to crop the image to the stage size
- add height and maxheight params
- set max width as 100% by default


src/tags/object/Image.js
```diff
@@ -266,6 +268,9 @@ const Model = types.model({
 }).volatile(() => ({
   currentImage: 0,
   stageRatio: 1,
+  containerWidth: 1,
+  containerHeight: 1,
+  targetWidth: 0,
```
hlomzik, line 273: not used

src/tags/object/Image.js
```diff
@@ -266,6 +268,9 @@ const Model = types.model({
 }).volatile(() => ({
   currentImage: 0,
   stageRatio: 1,
+  containerWidth: 1,
```
hlomzik, line 271: better to add comment that these params are used to calculate scale factor

src/components/ImageView/ImageView.js
```diff
@@ -613,55 +608,10 @@ export default observer(
...
-        }
-      }
-
-      if (imgTransform?.length > 0) {
-        imgStyle.transform = imgTransform.join(" ");
-      }
-
       if (this.imageRef.current) {
-        Object.assign(this.imageRef.current.style, imgStyle);
+        Object.assign(this.imageRef.current.style, item.imageTransform);
```
hlomzik, line 612: we update styles in two places (almost) simultaneously — here and in `style` prop during render, so check do we really need to do this twice. this change will happen before render, and that one — after.

hlomzik (approved):

hlomzik (approved):

PR #412, opened 2021-12-28
DEV-812: Selecting text inside HTML NER

Gondragos:


src/tags/object/RichText/view.js
```diff
@@ -363,15 +363,19 @@ class RichTextPieceView extends Component {
 
     // fix unselectable links
     const style = doc.createElement("style");
-
-    style.textContent = "body a { pointer-events: all; }";
+    const script = doc.createElement("script");
+
+    style.textContent = "body a { pointer-events: all !important; } .htx-highlight {display: inline !important;}";
+    script.textContent = `
+    document.querySelectorAll("a").forEach(a => a.removeAttribute("href"));
```
hlomzik, line 370: Removing `href` is destructive, especially when there is `clickableLinks=true` param.
Possible solution: move `href` to `data-href` and when user clicks on it we can open new tab programmatically. Also we can add `title` with link url to show it to user.

src/tags/object/RichText/view.js
```diff
@@ -363,15 +363,19 @@ class RichTextPieceView extends Component {
...
+    window.addEventListener("mousedown", (ev)=>{
```
hlomzik, line 371: why?

hlomzik (changes requested):

hlomzik:
For now this approach should be revised and priority of the bug is low, so closing PR.

PR #429, opened 2022-01-14
DEV-1264: Timeline sync

nicholasrq:


src/components/Timeline/Timeline.tsx
```diff
@@ -65,6 +65,21 @@ export const Timeline: FC<TimelineProps> = ({
...
+    setSeekViewPosition(newPosition);
+  };
+
+  const setSeekViewPosition = (newPosition: number) => {
+    const increase = newPosition > currentPosition;
+    const isInView = newPosition >= seekOffset && newPosition <= (seekOffset + seekVisibleWidth);
+
+    if (isInView) return;
+
+    if (increase && newPosition > (seekOffset + seekVisibleWidth)) {
```
hlomzik, line 78: if `currentPosition` is in current view then the only check we need is `if (increase)`

hlomzik (changes requested):

hlomzik (approved):

PR #459, opened 2022-02-01
fix: DEV-23: show line numbers again for Text

hlomzik:
Show them again if such setting enabled.
Uses `data-linenumbers` attribute to store current setting.

Also fix region items appearance in sidebar list.

### DIFF
--- a/src/stores/SettingsStore.js
+++ b/src/stores/SettingsStore.js
@@ -119,7 +119,8 @@ const SettingsModel = types
 
       // hack to enable it from outside, because Text spawns spans on every rerender
       // @todo it should be enabled inside Text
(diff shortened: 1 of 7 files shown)


src/stores/SettingsStore.js
```diff
@@ -119,7 +119,8 @@ const SettingsModel = types
 
       // hack to enable it from outside, because Text spawns spans on every rerender
       // @todo it should be enabled inside Text
-      document.querySelectorAll(".htx-text").forEach(text => text.classList.toggle("htx-line-numbers"));
+      document.querySelectorAll(".lsf-htx-richtext")
+        .forEach(text => text.dataset.linenumbers = self.showLineNumbers ? "enabled" : "disabled");
```
Gondragos, line 123: It seems that we do not need this part anymore

Gondragos (approved):

PR #461, opened 2022-02-01
fix: DEV-1564: DEV-1565: Fix shortcuts focus and cursor problems

Gondragos:


hlomzik (approved):

hlomzik:
Fixed in https://github.com/HumanSignal/label-studio-frontend/pull/462

PR #471, opened 2022-02-06
fix: DEV-1567: virtualize tree data

Travis1282:
Used react-window and react-vtree which is built on top, to virtualize expandable tree data in a variable height and width drop down taxonomy selector. The tree data viewing component is separate because it is agnostic of taxonomy\ and can be reused if needed in the future.


src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -307,7 +346,18 @@ const TaxonomyDropdown = ({ show, flatten, items, dropdownRef }: TaxonomyDropdow
         onInput={onInput}
         ref={inputRef}
       />
-      {list.map(item => <Item key={item.label} item={item} isFiltering={search !== ""} />)}
+      {show && (
+        <TreeStructure 
+          items={list} 
+          rowComponent={Item} 
+          flatten={search !== ""}
```
hlomzik, line 353: search results should be in a tree-like view, so we don't need this parameter anymore

src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -307,7 +346,18 @@ const TaxonomyDropdown = ({ show, flatten, items, dropdownRef }: TaxonomyDropdow
...
         ref={inputRef}
       />
-      {list.map(item => <Item key={item.label} item={item} isFiltering={search !== ""} />)}
+      {show && (
+        <TreeStructure 
+          items={list} 
+          rowComponent={Item} 
+          flatten={search !== ""} 
+          rowHeight={30}
+          defaultExpanded={true}
```
hlomzik, line 355: they should be collapsed by default, but expanded in search mode

src/components/TreeStructure/TreeStructure.tsx
```diff
@@ -0,0 +1,175 @@
...
+  }
+  return counter;
+};
+
+const blankItem: RowItem = ({ label: '', depth: 0, path: [], isOpen: true });
+
+
+const TreeStructure =  (
+  { items, rowComponent, flatten, rowHeight, maxHeightPersentage, minWidth, transformationCallback, defaultExpanded }: 
+  { items: any[],
```
hlomzik, line 68: so this should be `RowItem`?

src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -224,7 +243,9 @@ const Item = ({ item, isFiltering }: { item: TaxonomyItem, isFiltering?: boolean
           </div>
         )}
       </div>
-      {childs}
+      {isAdding && onAddLabel &&
```
hlomzik, line 246: `isFiltering` and comment about `key` is missing

src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -112,21 +135,27 @@ function isSubArray(item: string[], parent: string[]) {
...
   const { leafsOnly, maxUsages, maxUsagesReached, onAddLabel, onDeleteLabel } = useContext(TaxonomyOptionsContext);
-
-  const checked = selected.some(current => isArraysEqual(current, item.path));
-  const isChildSelected = selected.some(current => isSubArray(current, item.path));
-  const hasChilds = Boolean(item.children?.length);
+  
+  const checked = selected.some(current => isArraysEqual(current, path));
+  const isChildSelected = selected.some(current => isSubArray(current, path));
+  const hasChilds = !isLeaf;
   const onlyLeafsAllowed = leafsOnly && hasChilds;
```
hlomzik, line 151: so we can get rid of `hasChilds` and use even more readable `!isLeaf` directly

src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -183,15 +203,14 @@ const Item = ({ item, isFiltering }: { item: TaxonomyItem, isFiltering?: boolean
...
           />
-          {item.label}
+          {name}
         </label>
         {!isFiltering && (
           <div className={styles.taxonomy__extra}>
-            {/* @todo should count all the nested children */}
             <span className={styles.taxonomy__extra_count}>
-              {item.children?.length}
+              {childCount}
```
hlomzik, line 213: cool! that's the count of all the descendants?

src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -147,32 +176,23 @@ const Item = ({ item, isFiltering }: { item: TaxonomyItem, isFiltering?: boolean
...
-  }
 
   return (
-    <div>
+    <div style={{ paddingLeft: padding, ...style }}>
       <div className={[styles.taxonomy__item, customClassname].join(" ")}>
-        <div className={styles.taxonomy__grouping} onClick={toggle}>
-          <LsChevron stroke="#09f" style={arrowStyle} />
-        </div>
+        { !isFiltering && (
```
hlomzik, line 191: search results are also tree-like, so no conditions

src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -112,21 +135,27 @@ function isSubArray(item: string[], parent: string[]) {
...
+  const hasChilds = !isLeaf;
   const onlyLeafsAllowed = leafsOnly && hasChilds;
   const limitReached = maxUsagesReached && !checked;
   const disabled = onlyLeafsAllowed || limitReached;
 
-  const [isOpen, open, , toggle] = useToggle(isChildSelected || isFiltering);
-  const onClick = () => leafsOnly && toggle();
-  const arrowStyle = item.children?.length
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [_, open] = useToggle(isChildSelected || isFiltering);
```
hlomzik, line 156: it was used to open branches with selected items or when the first child is about to add, so now it useless :(
(also in search mode, but this can be fixed by `defaultExpanded`)

src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -112,21 +135,27 @@ function isSubArray(item: string[], parent: string[]) {
   return parent.every((n, i) => item[i] === n);
 }
 
+
 // @todo change `flat` into `alwaysOpen` and move it to context
-const Item = ({ item, isFiltering }: { item: TaxonomyItem, isFiltering?: boolean }) => {
+const Item: React.FC<any> = (props: RowProps) => {
```
hlomzik, line 140: first parameter of `React.FC` generic is for props, so it should be like `Item: React.FC<RowProps> = (props) => ...`

src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -112,21 +135,27 @@ function isSubArray(item: string[], parent: string[]) {
   return parent.every((n, i) => item[i] === n);
 }
 
+
 // @todo change `flat` into `alwaysOpen` and move it to context
```
hlomzik, line 139: could you please remove this redundant comment? I forgot to remove it

src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -183,15 +203,14 @@ const Item = ({ item, isFiltering }: { item: TaxonomyItem, isFiltering?: boolean
             disabled={disabled}
             checked={checked}
             ref={setIndeterminate}
-            onChange={e => setSelected(item.path, e.currentTarget.checked)}
+            onChange={e => setSelected(path, e.currentTarget.checked)}
           />
-          {item.label}
+          {name}
         </label>
         {!isFiltering && (
```
hlomzik, line 210: I see that this menu doesn't hide in search mode, so maybe that's because reconciliation doesn't treat this Item as changed, that's strange...

hlomzik (dismissed):

src/components/Taxonomy/Taxonomy.tsx
Travis1282: sorry :/

src/components/Taxonomy/Taxonomy.tsx
Travis1282: YUS!!

makseq (approved):

PR #473, opened 2022-02-07
fix: DEV-1555: RichText regions initialization

hlomzik:
- fix draft on HyperText loading. again. (wrap init with side-effects into freeze-and-replace history action)
- init all the offsets, range and text during init in one place
- use refs only from RichText model for the only source of truth
- reduce globalOffsets priority (first of all try to detect region by XPath in two ways — based on the original document and on labeled one, only then try to use globalOffsets)
- use globalOffsets internally for every highlight (or cached range)
- remove useless code (@todo which one?)


src/mixins/HighlightMixin.js
```diff
@@ -81,9 +85,8 @@ export const HighlightMixin = types
 
         selection.removeAllRanges();
         selection.addRange(range);
-        self._highlightedText = String(selection);
+        self.text = String(selection);
```
hlomzik, line 88: remove _highlightedText in volatile definition

src/tags/object/RichText/model.js
```diff
@@ -106,18 +106,16 @@ const Model = types
...
-    },
-
     get isReady() {
       return self.isLoaded  && self._isReady;
     },
   }))
   .volatile(() => ({
-    rootNodeRef: React.createRef(),
-    originalContentRef: React.createRef(),
+    rootNodeRef: React.createRef(), // outdated
```
hlomzik, line 114: remove this ref and describe other refs and all this logic

src/regions/RichTextRegion.js
```diff
@@ -112,42 +114,93 @@ const Model = types
...
+      return undefined;
+    },
+
+    // fix XPaths when we found region by globalOffsets
+    _fixXPaths(range, root) {
+      const normedRange = xpath.fromRange(range, root);
+
+      if (!isDefined(normedRange)) return;
+
+      self.start = normedRange.start ?? self.start;
```
hlomzik, line 200: why do we need this fallback to current values?
when it's possible that range is available, but some param of it is missing?

src/mixins/HighlightMixin.js
```diff
@@ -70,8 +74,8 @@ export const HighlightMixin = types
 
     updateHighlightedText() {
```
hlomzik, line 75: @todo remove this method

src/tags/object/RichText/model.js
```diff
@@ -232,40 +217,39 @@ const Model = types
...
+        }
+
         if (self.isLoaded === false) return;
         self.setReady(false);
         beforeNeedsUpdateCalback?.();
         self.regs.forEach(region => {
           try {
+            if (initial) {
+              region.initRangeAndOffsets();
+              region.updateHighlightedText();
```
hlomzik, line 234: check that text in ranges in hidden iframe are still accessible

src/regions/RichTextRegion.js
```diff
@@ -112,42 +114,93 @@ const Model = types
       });
     },
 
-    rangeFromGlobalOffset() {
-      const root = self._getRootNode();
+    getRangeToHighlight() {
+      if (!self.globalOffsets) return undefined;
```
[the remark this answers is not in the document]
hlomzik: add check for `root`

src/regions/RichTextRegion.js
```diff
@@ -112,42 +114,93 @@ const Model = types
...
+      }
+
+      // 3. if xpaths are broken use globalOffsets if given
+      if (self.globalOffsets && isDefined(root)) {
+        const { start, end } = self.globalOffsets;
+
+        self.globalOffsets = { start, end };
+        self.cachedRange = findRangeNative(start, end, root);
+
+        if (self.cachedRange) self._fixXPaths();
```
hlomzik, line 184: missed parameters

src/regions/RichTextRegion.js
```diff
@@ -112,42 +114,93 @@ const Model = types
...
+        self.cachedRange = range;
+
+        return;
+      }
+
+      // 3. if xpaths are broken use globalOffsets if given
+      if (self.globalOffsets && isDefined(root)) {
+        const { start, end } = self.globalOffsets;
+
+        self.globalOffsets = { start, end };
```
hlomzik, line 181: lolwut

hlomzik:
things to check:
- performance change (check master and current version on tons of regions)
- text retrieval for saveTextResult=false from hidden iframe
- init regions from auto-suggest (they might not have globalOffsets)

Gondragos (approved):

nicholasrq (approved):

PR #480, opened 2022-02-09
feat: DEV-24: Taxonomy styles and filtering

hlomzik:
Update Taxonomy design.
Keep tree structure even after filtering.


hlomzik:
fixed in #483

PR #507, opened 2022-02-25
feat: DEV-1656: Fix OCR Focus Text

superdev808:


examples/image_bbox/config.xml
```diff
@@ -1,7 +1,18 @@
...
+  <Polygon name="poly" toName="image" strokeWidth="3"/>
+  <TextArea name="transcription" toName="image"
+    editable="true"
+    perRegion="true"
+    required="true"
+    maxSubmissions="1"
+    rows="5"
+    placeholder="Recognized Text"
+    displayMode="region-list"
+  />
```
Gondragos, line 17: I guess in this case we need to create another example instead of editing the existent one.

Gondragos (approved):

PR #519, opened 2022-03-07
feat: DEV-1806: Fix Taxonomy TextArea Issue

superdev808:


src/core/Tree.tsx
```diff
@@ -236,7 +236,10 @@ function renderItem(el: IAnyStateTreeNode, includeKey = true) {
  */
 function renderChildren(item: IAnyStateTreeNode) {
   if (item && item.children && item.children.length) {
+    const defaultToName = item.children[0].isObject ? item.children[0].name : item.toname;
```
hlomzik, line 239: we should look for the first object in all the config, not in one level of controls we can fix. also this `toname` will be `undefined` in most of the cases (when the first tag is `Header` for example), so all the controls inside this `item` will be ruined.

src/tags/control/Base.js
```diff
@@ -1,26 +1,33 @@
...
-    return self.type;
-  },
+const ControlBase = types
+  .model({
+    smart: true,
+    smartonly: false,
+  })
+  .actions(self => ({
+    fixMissedToName(newToName) {
+      self.toname = newToName;
```
hlomzik, line 10: have to check is it defined

hlomzik (changes requested):

hlomzik:
Fixed in #505

PR #551, opened 2022-03-30
fix: dev-2057: width is auto adjustable 

Travis1282:
calculate row width based on the biggest visible taxonomy item

### DIFF
--- a/src/components/Taxonomy/Taxonomy.tsx
+++ b/src/components/Taxonomy/Taxonomy.tsx
@@ -56,10 +56,7 @@ type UserLabelFormProps = {
 
 interface RowProps {
   style: any;
-  index: number;
-  data: (
-    index: number,
-  ) => {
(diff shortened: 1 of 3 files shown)


src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -349,7 +350,8 @@ const TaxonomyDropdown = ({ show, flatten, items, dropdownRef }: TaxonomyDropdow
           rowHeight={30}
           defaultExpanded={false}
           maxHeightPercentage={50}
-          minWidth={300}
+          minWidth={200}
+          maxWidth={600}
```
hlomzik, line 354: would be better to pass these params to Taxonomy

hlomzik (approved):

hlomzik (approved):

PR #601, opened 2022-05-03
fix: DEV-2287: Disable drafts when history item is selected

nicholasrq:


hlomzik:
we'll be done in DEV-2290

PR #603, opened 2022-05-04
Correct the UNPKG.com url

Rafeethu:
the unpkg.com was referring the old pkg. corrected it to point to new package


hlomzik:
Thank you! We released new version and fixed unpkg link there!

PR #625, opened 2022-05-16
feat: DEV-2132: Bounding box annotation UX improvements

yyassi-heartex:
rotating bounding box feature


nicholasrq (changes requested):
See DEV-2131 for more details

PR #626, opened 2022-05-16
fix: dev-2068: pan-does-not-finsih-polygon

Travis1282:
check if is pan on toolswich from polygon, igonre if true


bmartel (approved):
Looks good to me.

Gondragos (approved):

PR #634, opened 2022-05-20
fix: DEV-1975: Taxonomy multi line items

Travis1282:
add support for super long taxonomy text wrapping multiple lines


yyassi-heartex (approved):
changes are good, but at some point we should stanardize either towards .styl or .scss but this isnt part of this PR or Jira Story

PR #639, opened 2022-05-25
feat: DEV-2340: UX Annotation improvement: Always show crosshairs when hovered on image viewport

yyassi-heartex:
this'll enable crosshairs so it'll be easier to see where the cursor is when it is hovered on the image viewport


yyassi-heartex:
I dont think we actually need this as we can toggle crosshair through settings

PR #651, opened 2022-05-30
fix: DEV-2351: Hypertext-security

juliosgarbi:
- I've changed the regex that remove scripts.
- I've added a new regex to remove all properties that starts with "on".


yyassi-heartex (approved):

juliosgarbi:
@yyassi-heartex as I said to Salvador, it's already impossible to add broke tags like what he said. That method is good if we could add broke tags like <scr<scr<script>something</script>ipt>Ha!</script>ipt> alert(document.cookie);</script> because if we clear the first script it will still print the second script.

yyassi-heartex (approved):

PR #665, opened 2022-06-01
fix: DEV-2465: Paragraphs data is not loading to QuickView

yyassi-heartex:
we were getting an array instead of a string, which caused a js error that prevents the page from working correctly.
solution: I now check if `str` is actually a string before continuing


bmartel (approved):
Looks good to me! I think the title of this should have been `fix: DEV-2465: Paragraphs data is not loading to QuickView` as this appears to be a bug not a new feature.

yyassi-heartex:
> Looks good to me! I think the title of this should have been `fix: DEV-2465: Paragraphs data is not loading to QuickView` as this appears to be a bug not a new feature.

Good catch! i've gone ahead and fixed it

Gondragos (approved):

PR #666, opened 2022-06-01
fix: DEV-2497: Enable smoothing

juliosgarbi:


src/tags/object/Image.js
```diff
@@ -57,6 +58,7 @@ const TagAttrs = types.model({
   height: types.maybeNull(types.string),
   maxwidth: types.optional(types.string, "100%"),
   maxheight: types.optional(types.string, "calc(100vh - 194px)"),
+  smoothing: types.maybeNull(types.boolean),
```
bmartel, line 61: Is this not an optional with a default as is stated in the jsdoc? I think it would make the logic a bit more clear if it was as the above typecheck is not required in this case.

src/stores/SettingsStore.js
```diff
@@ -58,6 +58,8 @@ const SettingsModel = types
     // showScore: types.optional(types.boolean, false),
 
     preserveSelectedTool: types.optional(types.boolean, true),
+
+    enableSmoothing:types.optional(types.boolean, true),
```
bmartel, line 62: spacing 

```js
enableSmoothing: types
```

src/components/ImageView/ImageView.js
```diff
@@ -454,6 +454,13 @@ export default observer(
     imageRef = createRef();
     crosshairRef = createRef();
 
+    constructor(props) {
+      super(props);
+
+      if(typeof props.item.smoothing === 'boolean')
```
bmartel, line 460: See my below comment about the optional with default instead of maybeNull, this could be then simplified to and no type check required:

```js
props.store.settings.setSmoothing(props.item.smoothing);
```

src/utils/feature-flags.ts
```diff
@@ -61,6 +61,7 @@ export const FF_DEV_2458 = "ff_front_dev_2458_comments_for_skip_250522_short";
 
 export const FF_DEV_2480 = "ff_dev_2480_convenient_offsets_270522_short";
 
+
```
bmartel, line 64: Is this space required here, or is it a leftover from some other work?

bmartel (changes requested):

src/stores/SettingsStore.js
juliosgarbi: @bmartel here I'm using the pattern of this document

src/tags/object/Image.js
hlomzik: we have similar behaviour for `showLabels` parameter and setting — if it's not set we use settings, if it's set we use this value from config. so actually it looks good, but jsdoc definition shouldn't have default value, and it's better to add that by default it uses user settings. @juliosgarbi

src/stores/SettingsStore.js
hlomzik: what do you mean by pattern?
we actually missed this spacing part in our linter, but here should be a space. and we definitely should cover this by linting rules.

hlomzik: so could you please add spaces here and in `if`s in `ImageView`?

src/components/ImageView/ImageView.js
```diff
@@ -741,6 +748,10 @@ export default observer(
         containerStyle["height"] = item.height;
       }
 
+      if(!this.props.store.settings.enableSmoothing && item.zoomScale > 1.4){
```
hlomzik, line 751: why 1.4? if we do this only for zoomed in images it should work right after 1

hlomzik (changes requested):

src/components/ImageView/ImageView.js
juliosgarbi: @hlomzik I was testing with 1.4 and I forgot to change to 1, I'll change! Thanks!!

bmartel (approved):
Looks good to me!

src/tags/object/Image.js
juliosgarbi: done!

src/stores/SettingsStore.js
juliosgarbi: done!

hlomzik (approved):

PR #675, opened 2022-06-02
fix: DEV-2462: Changing label

juliosgarbi:
- Resolved the problem that the first segment was mimicking the second segment.
- ClassList was duplicating if user changes the segment


bmartel (approved):
This is a weird bug, I was able to replicate the behaviour locally, and this does in fact fix the issue. Nice!

PR #684, opened 2022-06-07
fix: DEV-1422: deselection of ImageView bounding box before creating another

bmartel:


src/components/ImageView/ImageView.js
```diff
@@ -475,10 +475,28 @@ export default observer(
...
+      if (
+        isFF(FF_DEV_1442) &&
+        e.target === item.stageRef &&
+        item.annotation.selectedRegions.length > 0 &&
+        [
+          undefined,
+          "EllipseTool",
+          "EllipseTool-dynamic",
+          "RectangleTool",
+          "RectangleTool-dynamic",
```
nicholasrq, line 489: We also need this behavior for closed polygons

nicholasrq (changes requested):

src/components/ImageView/ImageView.js
bmartel: good point, let me add that in!

nicholasrq (approved):

PR #685, opened 2022-06-07
fix: DEV-1975: DEV-1843: Support long unbroken taxonomy line breaks

Travis1282:
In order to support text without spaces that exceed the maximum width for taxonomy, creation of an invisible taxonomy item to check its width and calculate how many lines was necessary per item. Additionally I added the missing visibility mixin check.


juliosgarbi:
Looks good to me!

juliosgarbi (approved):

PR #697, opened 2022-06-09
fix: DEV-2303: Video regions are editable when anno history is selected

nicholasrq:


src/tags/object/Video/Video.js
```diff
@@ -1,4 +1,4 @@
-import { getRoot, types } from "mobx-state-tree";
+import { getParent, getRoot, types } from "mobx-state-tree";
```
bmartel, line 1: `getParent` is unused

bmartel (changes requested):

PR #698, opened 2022-06-09
fix: DEV-2510: add guard with error message for non incremental, non sequential datasets. (#1752)

bmartel:
Timeseries data is required to be ordered by an incrementing sequence to work with the TimeSeries tag. This adds a check when mapping the timestamps to find an out of order or decrementing value in the series.

https://labelstud.io/tags/timeseries.html


src/tags/object/TimeSeries.js
```diff
@@ -160,7 +160,31 @@ const Model = types
...
+        let current = self.parseTime(data[self.keyColumn][0]);
+        const first = current;
+        const nonSeqValues = [`seq: 0, value: ${data[self.keyColumn][0]}`];
+
+        const timestamps = data[self.keyColumn].map((value, i) => {
+          // Compare previous and current timestamp to determine direction is ascending as the values must be sorted in ascending order to be valid timeseries data
+          if (i > 0) {
+            current = self.parseTime(value);
+            const diff = current - first;
+            const direction= diff > 0 ? 1 : -1;
```
makseq, line 172: `direction=` => `direction =`

src/tags/object/TimeSeries.js
```diff
@@ -160,7 +160,31 @@ const Model = types
...
-        const timestamps = data[self.keyColumn].map(self.parseTime);
+        let current = self.parseTime(data[self.keyColumn][0]);
+        const first = current;
+        const nonSeqValues = [`seq: 0, value: ${data[self.keyColumn][0]}`];
+
+        const timestamps = data[self.keyColumn].map((value, i) => {
+          // Compare previous and current timestamp to determine direction is ascending as the values must be sorted in ascending order to be valid timeseries data
+          if (i > 0) {
+            current = self.parseTime(value);
+            const diff = current - first;
```
makseq, line 171: Let's try this csv: 
```
time,value
1,0.1
3,0.2
2,0.5
5,0.8
4,0.7
```

makseq (changes requested):

makseq (approved):

PR #699, opened 2022-06-10
fix: DEV-1714: X-axis zoom threshold

Travis1282:
Increase the minimum and default zoom to 300, the min set for wave-surfer

### DIFF
--- a/src/components/Timeline/Views/Wave/Wave.tsx
+++ b/src/components/Timeline/Views/Wave/Wave.tsx
@@ -21,6 +21,7 @@ const ZOOM_X = {
   max: 1500,
   step: 10,
   default: 300,
+  defaultValue: 300,
 };
 
 const SPEED = {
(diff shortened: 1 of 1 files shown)


src/components/Timeline/Views/Wave/Wave.tsx
```diff
@@ -17,10 +17,10 @@ import { useMemo } from "react";
 import { WaveSurferParams } from "wavesurfer.js/types/params";
 
 const ZOOM_X = {
-  min: 10,
+  min: 300,
   max: 1500,
   step: 10,
-  default: 10,
+  default: 300,
```
bmartel, line 23: Seeing as these values are spread on the `Range`, would it not be better to change `default` to `defaultValue` and just use that instead? Seems more semantic in the usage, unless `default` is also operating as its own value within the component. I just find it unnerving when using keywords as attributes.

bmartel (changes requested):

src/components/Timeline/Views/Wave/Wave.tsx
Travis1282: Ah, I see what you mean. I think we need both though.

bmartel (approved):

PR #700, opened 2022-06-10
fix: DEV-2301: Fix classification type tags render on anno history change

nicholasrq:


src/tags/TagBase.js
```diff
@@ -0,0 +1,17 @@
+import { types } from "mobx-state-tree";
+import { isDefined } from "../utils/utilities";
+
+const BaseTag = types
+  .model("BaseTag")
+  .views((self) => ({
+    /**
+     * Identify classification type tags
+     *
+     * `perRegionVisible` is only available for classification type tags
```
bmartel, line 10: if this tag mixin is only related to `Classification` types, would it not be better to name this `ClassificationTag` or `ClassificationBaseTag`? I think it would be a bit more clear as the usage, as BaseTag is a bit overloaded of a term.

bmartel (approved):

PR #701, opened 2022-06-10
fix: DEV-2365: User is able to edit choice results imported as read-only

yyassi-heartex:
added new readonly flag to annotation based on if the result imported is readonly. if annotation is readonly then choices with in are also flagged as readonly


yyassi-heartex:
OOPS. wrong branch name. need to fix this

PR #702, opened 2022-06-10
fix: DEV-2365: User is able to edit choice results imported as read-only

yyassi-heartex:
Annotations now get a readonly flag based on the results that were imported.

if Annotation is readonly then Choices are readonly


src/stores/Annotation/Annotation.js
```diff
@@ -72,6 +72,8 @@ export const Annotation = types
     regionStore: types.optional(RegionStore, {
       regions: [],
     }),
+
+    readonly: types.optional(types.boolean, false),
```
bmartel, line 76: Previous to making this an explicit state (which I like by the way), how was an annotation considered implicitly `readonly`? 

I am asking for two reasons:
1. I could use this to fix the bug I am on right now very easily.
2. Does this work at all with non imported annotations, or how is that impacted?

bmartel: @nicholasrq ?

yyassi-heartex: as far as im aware, there was no implicit way to deal with it. I just saw that ImageModel (and so on) was getting the readonly flag and needed a way to make sure we can pass that down to children (Choices) properly.

1. if this route is approved then yes that sounds like a good solution to the problem you are trying to address.
2. i believe it would on non-imported annotations but i'm not sure how to created a readonly annotation with out importing it

bmartel (approved):

PR #703, opened 2022-06-10
fix: DEV-1833: Prop whenChoiceValue working as expected

juliosgarbi:


src/mixins/Visibility.js
```diff
@@ -73,6 +72,19 @@ const VisibilityMixin = types
 
           return res !== false;
         }
+      }else if (self.whenchoicevalue){
```
bmartel, line 75:
```suggestion
      } else if (self.whenchoicevalue) {
```

src/mixins/Visibility.js
```diff
@@ -73,6 +72,19 @@ const VisibilityMixin = types
...
+        for (const choices of self.annotation.names.values()) {
+          if (choices.selectedValues && choices.selectedValues().length){
```
bmartel, line 77:
```suggestion
          const choicesList = choices?.selectedValues?.()
          if (choicesList?.length) {
```

src/mixins/Visibility.js
```diff
@@ -73,6 +72,19 @@ const VisibilityMixin = types
...
+            const choicesList = choices.selectedValues();
```
bmartel, line 78:
```suggestion
```

bmartel (changes requested):

Travis1282 (approved):

PR #705, opened 2022-06-13
fix: DEV-2592: Bounding box is flickering during region creation

yyassi-heartex:
this corrects an issue where the region flickers while dragging out a rectangle when using 3 point bbox feature


bmartel (approved):
Looks good to me

juliosgarbi (approved):

juliosgarbi:
Looks good!

PR #708, opened 2022-06-14
fix: DEV-2313: Unfinished polygon zoom fix

Travis1282:
Update drawing region sizes on zoom


juliosgarbi (approved):

src/tags/object/Image.js
```diff
@@ -836,6 +836,7 @@ const Model = types.model({
       self.regs.forEach(shape => {
         shape.updateImageSize(width / naturalWidth, height / naturalHeight, width, height, userResize);
       });
+      self.drawingRegion.updateImageSize(width / naturalWidth, height / naturalHeight, width, height, userResize);
```
Gondragos, line 839: `self.drawing` may be null

Gondragos (changes requested):

Gondragos (approved):

PR #709, opened 2022-06-15
fix: DEV-2207: sync speed between video and audio

juliosgarbi:
- fixed the bug video and audio wasn't sync when play speed was changed


bmartel (approved):
Looks good to me!

nicholasrq (approved):

PR #713, opened 2022-06-17
fix: DEV-117: DEV-1179: DateTime tag, fixes for date save and region switch

Travis1282:
Date change on region click and date not saved fixes


src/tags/control/DateTime.js
```diff
@@ -0,0 +1,366 @@
...
+  observer(({ item }) => {
+    const visibleStyle = item.perRegionVisible() ? { margin: "0 0 1em" } : { display: "none" };
+    const visual = {
+      style: { width: "auto", marginRight: "4px" },
+      className: "ant-input",
+    };
+    const [minTime, maxTime] = [item.min, item.max].map(s => s?.match(/\d?\d:\d\d/)?.[0]);
+    const [dateInputValue, setDateInputValue] = useState("");
+
+    const hanldeDateInputValueChange = event => {
```
bmartel, line 300: typo
```suggestion
    const handleDateInputValueChange = event => {
```

src/tags/control/DateTime.js
```diff
@@ -0,0 +1,366 @@
...
+        )}
+        {item.showDate && (
+          <input
+            {...visual}
+            type="date"
+            name={item.name + "-date"}
+            value={dateInputValue}
+            min={item.min}
+            max={item.max}
+            onChange={hanldeDateInputValueChange}
```
bmartel, line 345: typo
```suggestion
            onChange={handleDateInputValueChange}
```

bmartel (changes requested):

src/tags/control/DateTime.js
```diff
@@ -198,6 +203,15 @@ const Model = types
       self.time = undefined;
     },
 
+    validDateFormat(dateString) {
+      const dateNumberArray = dateString.split("-").map(dateString => parseInt(dateString, 10));
+      const year = dateNumberArray[0];
+      const jsDate = new Date(dateString);
+
+      if (isNaN(jsDate) || year < 1000 || year >= 10000) return false;
```
bmartel, line 211: What range of date years are supported for this? Can I as a user enter a date in the past or future? If so, why the valid year range of 1001-9999?

src/tags/control/DateTime.js
Travis1282: looks my spell checker went kaput

src/tags/control/DateTime.js
Travis1282: there are valid dates that are not valid to our system the year must be four positive integers and this is to check that, I will make more readable.

bmartel (approved):
Looks good to me!

nicholasrq (changes requested):
I assume that new feature should be covered with tests right away, let's do it.

I would also love to see an example added to `examples/` dir for everyone to play with it.

hlomzik (approved):

PR #716, opened 2022-06-21
fix: DEV-2432: Allow unfinished polygons to be saved as draft

bmartel:


juliosgarbi (approved):
looks good to me

src/mixins/DrawingTool.js
```diff
@@ -154,7 +155,16 @@ const DrawingTool = types
       startDrawing(x, y) {
         self.annotation.history.freeze();
         self.mode = "drawing";
-        const currentArea = self.createDrawingRegion(self.createRegionOptions({ x, y }));
+
+        const isPolygon = self.control.type.startsWith("polygon");
```
yyassi-heartex, line 159: the code here is good, my only small recommendation would be to create startDrawing in editor/src//tools/Polygon.js and extend the DrawingTool.startDrawing.
this way you dont need to do the check for isPolygon and just do a check for the FF

yyassi-heartex (commented):
i think its pretty good, just a small recommendation - let me know your thought on this

src/mixins/DrawingTool.js
bmartel: The reason why I think that is the better way to go is because it properly encapsulates the behaviour difference in the correct location. Good suggestion 👍

PR #717, opened 2022-06-21
fix: DEV-2655: Allow cmd or ctrl to create regions immediately with deselect on canvas click enabled.

bmartel:


nicholasrq (approved):

yyassi-heartex (approved):
LGTM

PR #737, opened 2022-07-06
fix: DEV-2806: Unable to delete relationships between regions

yyassi-heartex:


bmartel:
Closing due to process change, code already present in master.

PR #745, opened 2022-07-12
fix: DEV-2431: Improve delete point logic

Travis1282:
remove problematic logic in delete point function


bmartel (approved):
Verified changes on deployment.

PR #749, opened 2022-07-13
fix: DEV-2432: Allow polygon to save closed state as snapshot to retain state of unfinished polygon points.

bmartel:


bmartel:
Closing as is wrong ticket number.

PR #750, opened 2022-07-13
fix: DEV-2671: Change rotate anchor of bbox

juliosgarbi:
- Fix rotation position
- fix bbox rotating outisde of boundaries


yyassi-heartex (changes requested):
there are some minor tweaks, but nothing too crazy - its very close

hlomzik (changes requested):
I rewrote calculations and it was easier to re-create PR: #768

PR #751, opened 2022-07-13
fix: DEV-2432: Allow polygon to save closed state as snapshot to retain state of unfinished polygon points

bmartel:


juliosgarbi (approved):

src/stores/Annotation/Annotation.js
```diff
@@ -761,7 +761,15 @@ export const Annotation = types
       document.body.style.cursor = "wait";
 
       const result = self.results
-        .map(r => r.serialize(options))
+        .map(r => {
+          const serialized = r.serialize(options);
+
+          if (isFF(FF_DEV_2432) && r.type === "polygonlabels") {
+            serialized.closed = r.area?.closed;
```
hlomzik, line 768: it's better to move this to PolygonRegion#serialize to keep it clean

hlomzik (changes requested):

yyassi-heartex (approved):
LGTM

e2e/fragments/LabelStudio.js
```diff
@@ -4,8 +4,14 @@ const { I } = inject();
 const Helpers = require("../tests/helpers");
 
 module.exports = {
-  init(params) {
-    I.executeScript(Helpers.initLabelStudio, params);
+  init({ events = {}, ...params }) {
```
bmartel, line 7: Maybe add a comment here about the difference about the `events` binding vs the dynamic function injections we are using at runtime.

src/mixins/DrawingTool.js
```diff
@@ -113,13 +113,21 @@ const DrawingTool = types
       createDrawingRegion(opts) {
         const control = self.control;
         const resultValue = control.getResultValue();
-
+
```
bmartel, line 116:
```suggestion
```

bmartel (commented):
Unclosed polygons history should work after draft save with hotkeys (local history).

nicholasrq (approved):

PR #754, opened 2022-07-15
fix: DEV-2696: 3point bbox tool remains selected but hidden, when user changes label configuration from Rectangle to RectangleLabels

yyassi-heartex:
resolving an issue where 3 point rectangles were drawing before selecting a different tool even after removing them as an option from labeling settings


juliosgarbi (approved):
LGTM!

PR #761, opened 2022-07-20
feat: DEV-2755: Regions list grouped by labels with persistent & ordered collapse

yyassi-heartex:


juliosgarbi (changes requested):
Features must be behind feature flags also there is some inconsistency in the spacing between '( )' some places have some not

juliosgarbi (dismissed):
LGTM!

juliosgarbi (approved):
LGTM

src/components/Entities/LabelList.js
```diff
@@ -19,6 +43,8 @@ export const LabelList = observer(({ regionStore }) => {
     };
   });
 
+  const expandedKeys = treeData.filter( item => !collapsedPos.includes( item.pos ) ).map( item => item.key ) ?? [];
```
bmartel, line 46: probably should put any of the new additional functionality behind the FF.

the `Tree` props `expandedKeys` and `onExpand` will have the new behaviour by default right now.

bmartel (changes requested):

bmartel (approved):

PR #772, opened 2022-07-22
fix: DEV-2934: Fix Audio resizing behavior

nicholasrq:
Audio doesn't properly react to the layout resize. It causes regions to be shifted when resize happens when  zoom is 1.


src/components/Timeline/Views/Wave/Wave.tsx
```diff
@@ -242,6 +258,26 @@ export const Wave: FC<TimelineViewProps> = ({
     }
   }, [scale]);
 
+  useEffect(() => {
+    const observer = new ResizeObserver(() => {
+      const wsi = ws.current;
+
+      requestAnimationFrame(() => {
+        // if (wsi) wsi.zoom(storedPosition.current.zoom);
```
bmartel, line 266: forgot some commented out code here

bmartel (changes requested):
looks great, just a line of code left over

bmartel (approved):

hlomzik (approved):

PR #777, opened 2022-07-25
fix: DEV-2395: Quick view: Different size of working area is created for SMALL vertical/horizontal images

yyassi-heartex:
this will make the view container always fit the maximum available space in the viewport


src/tags/visual/View.js
```diff
@@ -110,8 +111,15 @@ const HtxView = observer(({ item }) => {
     style["display"] = "none";
   }
 
+  const className = [
+    item.className,
+    styles.viewContainer,
+  ]
+    .filter((className) => !!className)
```
bmartel, line 118:
```suggestion
    .filter(Boolean)
```

bmartel (approved):
Looks good to me, added a small suggestion.

src/tags/visual/View.js
```diff
@@ -110,8 +112,15 @@ const HtxView = observer(({ item }) => {
     style["display"] = "none";
   }
 
+  const className = [
+    item.className,
+    styles.viewContainer,
```
bmartel, line 117: since its already setup to do the conditional className building, why not just do this instead:

```tsx
isFF(FF_DEV_2395) && styles.viewContainer
```

src/tags/visual/View.js
```diff
@@ -110,8 +112,15 @@ const HtxView = observer(({ item }) => {
...
+  ]
+    .filter(Boolean)
+    .join(" ");
+
   return (
-    <div className={item.classname} style={style}>
+    <div className={isFF(FF_DEV_2395) ? className : item.className} style={style}>
```
bmartel, line 123: With that you can just leave the `className={item.className}`

bmartel (changes requested):

src/tags/visual/View.js
yyassi-heartex: done

src/tags/visual/View.js
yyassi-heartex: done

bmartel (approved):
looks good to me

yyassi-heartex:
closing this PR as it is no longer needed

PR #778, opened 2022-07-25
fix: DEV-2576: Undo keypoints when create new polygon

juliosgarbi:


src/regions/PolygonRegion.js
```diff
@@ -65,225 +65,239 @@ const Model = types
...
 
-    closePoly() {
-      self.closed = true;
-    },
+        self.points.splice(insertIdx, 0, p);
+      },
 
-    canClose(x, y) {
-      if (self.points.length < 2) return false;
+      undoRedoPoints(undo = true){
```
hlomzik, line 174: spacing

src/mixins/DrawingTool.js
```diff
@@ -3,6 +3,14 @@ import { types } from "mobx-state-tree";
 import Utils from "../utils";
 import throttle from "lodash.throttle";
 import { MIN_SIZE } from "../tools/Base";
+import { Hotkey } from "../core/Hotkey";
+
+const initializeHotkeys = (self) =>{
+  const hotkeys = Hotkey("Polygons");
+
+  hotkeys.addNamed("polygon:undo", () => self.getCurrentArea()?.undoRedoPoints(true));
+  hotkeys.addNamed("polygon:redo", () => self.getCurrentArea()?.undoRedoPoints(false));
```
hlomzik, line 12: you should always remove hotkeys on destory, [check Tool.js](https://github.com/heartexlabs/label-studio-frontend/blob/master/src/components/Toolbar/Tool.js#L82-L100) for example or search for `hotkey.` usages in the project.

src/regions/PolygonRegion.js
hlomzik, line 174: there are no common logic, so it would be more clear to split this method in two: undoPoint/redoPoint

src/stores/AppStore.js
```diff
@@ -338,13 +338,13 @@ export default types
       hotkeys.addNamed("annotation:undo", function() {
         const annotation = self.annotationStore.selected;
 
-        annotation.undo();
+        if (!annotation.isDrawing) annotation.undo();
```
hlomzik, line 341: actually these two lines can be very helpful!

hlomzik (changes requested):

Travis1282 (approved):

hlomzik (changes requested):
main question, as usual — feature flags? ;)

src/components/ImageView/ImageView.js
```diff
@@ -746,11 +746,20 @@ export default observer(
...
+      const annotation = store.annotationStore.selected;
+
       window.addEventListener("resize", this.onResize);
-      this.attachObserver(this.props.item.containerRef);
+      this.attachObserver(item.containerRef);
       this.updateReadyStatus();
 
       hotkeys.addDescription("shift", "Pan image");
+      hotkeys.addNamed("polygon:undo", () => {
+        if (annotation.isDrawing) item.drawingRegion.undoPoints();
```
hlomzik, line 758: `DrawingRegion` is used not only in polygons, so you can get an error in this line. but I think it would be enough to just add `?.`

src/regions/PolygonRegion.js
```diff
@@ -65,225 +65,240 @@ const Model = types
...
-    canClose(x, y) {
-      if (self.points.length < 2) return false;
+      undoPoints(){
+        this.deletePoint(self.points[self.points.length - 1]);
+      },
 
-      const p1 = self.points[0];
-      const p2 = { x, y };
+      redoPoints(){
+        if(_historyPoints[self.points.length])
```
hlomzik, line 179: fix spacing and add braces in this methods; also it's better to move `_historyPoints[self.points.length]` to a var

hlomzik (dismissed):

Travis1282 (approved):

PR #783, opened 2022-07-27
fix: DEV-3018: make all user references the same type so comments and other references are not made invalid

bmartel:


src/stores/AppStore.js
```diff
@@ -136,8 +136,16 @@ export default types
     userLabels: isFF(FF_DEV_1536) ? types.optional(UserLabels, { controls: {} }) : types.undefined,
   })
   .preProcessSnapshot((sn) => {
-    if (!sn.user) {
-      sn.user = window.APP_SETTINGS?.user ?? null;
+    if (isFF(FF_DEV_2887)) {
+      const currentUser = window.APP_SETTINGS?.user ?? null;
```
hlomzik, line 140: user can also be in `sn.user`, so I'd put it as the first option to check

src/stores/AppStore.js
```diff
@@ -136,8 +136,16 @@ export default types
...
+
+      if (sn.users?.length) {
+        sn.users = [currentUser, ...sn.users.filter(({ id }) => id !== currentUser.id)];
```
hlomzik, line 143: `currentUser` can be null

src/stores/AppStore.js
```diff
@@ -136,8 +136,16 @@ export default types
...
+      }
+
+      if (currentUser?.id) {
+        sn.user = currentUser.id;
```
hlomzik, line 147: we should always set `sn.user` otherwise it can break reference if user came from constructor

src/stores/Comment/CommentStore.js
```diff
@@ -78,7 +79,7 @@ export const CommentStore = types
       const toPersist = self.queuedComments;
 
       if (!self.canPersist || !toPersist.length) return;
-
+
```
hlomzik, line 82: accidential space

src/stores/Comment/CommentStore.js
```diff
@@ -152,11 +153,16 @@ export const CommentStore = types
...
-          if (queueRestored) {
-            restored.comments = restored.comments.map(comment => ({ id: comment.id > 0 ? comment.id * -1 : comment.id,  ...comment }));
+          let restoreIds = [];
+
+          if (queueRestored) { 
+            restoreIds = restored.comments.map(comment => comment.id);
           }
           if (merge) {
-            restored.comments = [...restored.comments, ...getSnapshot(self.comments)].sort((a, b) => new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime());
+            restored.comments = uniqBy([...restored.comments, ...getSnapshot(self.comments)], 'id').sort((a, b) => new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime());
```
hlomzik, line 162: line is too long (as well as the line in next `if`)

hlomzik (changes requested):

src/stores/AppStore.js
bmartel: I'll just always set sn.user to currentUser if currentUser exists, as its always correct (I had @makseq adjust this to make double sure its available with correct fields).

src/stores/AppStore.js
bmartel: I agree, see my other comment.

src/stores/AppStore.js
hlomzik: But you miss situation when `user` is passed via the params to LabelStudio constructor. And LSF can be used as standalone module as well.

hlomzik: https://github.com/heartexlabs/label-studio-frontend/blob/master/README.md?plain=1#L70-L74

src/stores/AppStore.js
```diff
@@ -136,8 +136,16 @@ export default types
     userLabels: isFF(FF_DEV_1536) ? types.optional(UserLabels, { controls: {} }) : types.undefined,
   })
   .preProcessSnapshot((sn) => {
-    if (!sn.user) {
-      sn.user = window.APP_SETTINGS?.user ?? null;
```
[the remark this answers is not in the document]
bmartel: Updated in the changes. Now considers both user's and has an overall check if neither are present.

hlomzik (approved):

src/stores/AppStore.js
```diff
@@ -136,8 +136,16 @@ export default types
...
+    const currentUser = window.APP_SETTINGS?.user ?? sn.user ?? null;
```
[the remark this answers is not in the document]
bmartel: agreed

PR #784, opened 2022-07-28
fix: DEV-3004: The Zoom setting changes "Accept" to "Fix + Accept" during Review Stream

yyassi-heartex:


yyassi-heartex:
closing this to resolving an error in which task this is linked to

PR #787, opened 2022-07-28
fix: DEV-3018: Convert app store user to reference only if the incoming value is not already a reference

bmartel:


bmartel:
@hlomzik There were issues I had found when leaving the sn.user as the default, it can be possible to already by the user id reference so I added a condition to only update the sn.user value if its not a number.

yyassi-heartex (approved):

PR #790, opened 2022-07-29
fix: DEV-3049: visual bug with the Zoom To flyout menu on smaller screen sizes

yyassi-heartex:
minor correction for a fly out menu item positioning incorrectly on smaller screens (smaller than 1200px wide)


juliosgarbi (approved):
LGTM

PR #801, opened 2022-08-05
fix: DEV-3110: Load draft comments correctly

bmartel:


src/stores/Comment/CommentStore.js
```diff
@@ -164,7 +164,8 @@ export const CommentStore = types
 
       if (self.annotationId) {
         comment.annotation = self.annotationId;
-      } else if (self.draftId) {
+      }
+      if (self.draftId) {
```
hlomzik, line 168: the logic here: if there is already submitted annotation, we should not attach comment to a draft, only to annotation directly.

hlomzik: but LSF is actually a standalone separated lib actually, so maybe it's better to do it like in your fix and move this conditional logic to api part (dm)

src/stores/AppStore.js
```diff
@@ -687,6 +688,14 @@ export default types
       }
     }
 
+    function setUsers(users) {
+      self.users.replace(users);
+    }
+
+    function mergeUsers(users) {
+      self.setUsers(uniqBy([...getSnapshot(self.users), ...users], 'id'));
```
hlomzik, line 696: what would happen if you just pass an array with non-unique users to mst? error or some merging policy? because maybe we can simplify it a bit.

hlomzik (changes requested):

hlomzik:
more like discussion than changes request

src/stores/AppStore.js
nicholasrq: it will cause an error with `types.identifierNumber` duplication

src/stores/Comment/CommentStore.js
hlomzik: oh, I see, in both places it sends correct ids to api!

hlomzik (approved):

PR #803, opened 2022-08-09
fix: DEV-3077: Repeater performance improvement

juliosgarbi:


yyassi-heartex (approved):
looks good to me - but this may exasperate some existing issues with how the image sizes on page load making it more difficult to fit to available space on page load (not related to this task)

PR #812, opened 2022-08-17
fix: DEV-3211: Parse all timeseries timestamps to utc to avoid improper calculations using possible daylight savings time

bmartel:


yyassi-heartex (approved):

src/tags/object/TimeSeries.js
```diff
@@ -141,13 +141,19 @@ const Model = types
     },
 
     get parseTimeFn() {
-      return self.timeformat && self.timecolumn ? d3.timeParse(self.timeformat) : Number;
+      return self.timeformat && self.timecolumn ? d3.utcParse(self.timeformat) : Number;
```
hlomzik, line 144: it requires more checks and fixes...
`utcParse` parses date as it was in UTC, but store it in current timezone:
```
d3.utcParse("%Y-%m-%d")("2022-08-19")
> Fri Aug 19 2022 02:00:00 GMT+0200 (Central European Summer Time)
```
not sure how it could affect the logic, we have to check how it works with `formatTrackerTime()`.
But with our current `formatTime()` it can mess the dates:
```
p=d3.utcParse("%Y-%m-%d %H")
f=d3.timeFormat("%Y-%m-%d %H")
f(p("2022-08-19 23"))
> '2022-08-20 01'
```
so at least we have to also fix `d3.timeFormat` to `d3.utcFormat`

hlomzik (changes requested):

src/tags/object/TimeSeries.js
bmartel: The more I think about this, the more it seems wrong to ever have localized times for graphing data. If I am charting sensor data from a manufacturing plant that operates in a region of the world, I would expect all the datetime values reported to be local to there, not converting to wherever I am reading this chart from. Because if I am told there is a particular range of time that is of importance by another user, its now incorrect whatever they told me because we are observing completely different values. We need this parsing and formatting to remain in UTC both sides of all timeseries related data for sure, otherwise this is incorrect to the users almost ubiquitously.

bmartel: Seems this works, it captures and shows correctly the original time without the issue of locale or daylight savings time.

Screen Shot 2022-08-19 at 11 20 37 AM

hlomzik (approved):

PR #818, opened 2022-08-24
feat: DEV-388: Introducing default attributes for AudioPlus

nicholasrq:
Introducing `defaultVolume`, `defaultSpeed` and `defaultZoom`


bmartel (approved):

src/tags/object/AudioNext/view.tsx
```diff
@@ -14,9 +14,9 @@ const AudioNextView: FC<AudioNextProps> = ({ item }) => {
   const [position, setPosition] = useState(1);
   const [audioLength, setAudioLength] = useState(0);
 
-  const [zoom, setZoom] = useState(WS_ZOOM_X.default);
-  const [volume, setVolume] = useState(1);
-  const [speed, setSpeed] = useState(1);
+  const [zoom, setZoom] = useState(Number(item.defaultzoom) ?? WS_ZOOM_X.default);
```
bmartel, line 17: Actually looking at these again this will never fallback to anything since `Number()` will produce a value not undefined or null. Probably need to split that out.

bmartel (changes requested):

src/tags/object/AudioNext/view.tsx
bmartel: something like 
```suggestion
  const [zoom, setZoom] = useState(Number(item.defaultzoom ?? WS_ZOOM_X.default));
```

src/tags/object/AudioNext/view.tsx
```diff
@@ -14,9 +14,9 @@ const AudioNextView: FC<AudioNextProps> = ({ item }) => {
...
+  const [volume, setVolume] = useState(Number(item.defaultvolume) ?? 1);
```
bmartel, line 18:
```suggestion
  const [volume, setVolume] = useState(Number(item.defaultvolume ?? 1));
```

src/tags/object/AudioNext/view.tsx
```diff
@@ -14,9 +14,9 @@ const AudioNextView: FC<AudioNextProps> = ({ item }) => {
...
+  const [speed, setSpeed] = useState(Number(item.defaultspeed) ?? 1);
```
bmartel, line 19:
```suggestion
  const [speed, setSpeed] = useState(Number(item.defaultspeed ?? 1));
```

src/tags/object/AudioNext/view_old.js
```diff
@@ -30,6 +30,9 @@ const HtxAudioView = ({ store, item }) => {
           onError={item.onError}
           speed={item.speed}
           zoom={item.zoom}
+          defaultVolume={Number(item.defaultvolume)}
+          deafultSpeed={Number(item.deafultspeed)}
```
bmartel, line 34:
```suggestion
          deafultSpeed={Number(item.defaultspeed)}
```

src/common/Utils/useValueTracker.ts
```diff
@@ -5,7 +5,7 @@ export const useValueTracker = <T = any>(
   defaultValue?: T | undefined,
 ): [T | string, (value: T) => void] => {
   const initialValue = useMemo(() => {
-    return defaultValue ?? value ?? "";
+    return (value || defaultValue) ?? "";
```
bmartel, line 8: is this value tracker ever used for booleans `false` or `0`, if so this might break expected behaviour.

src/tags/object/AudioNext/view_old.js
```diff
@@ -30,6 +30,9 @@ const HtxAudioView = ({ store, item }) => {
           onError={item.onError}
           speed={item.speed}
           zoom={item.zoom}
+          defaultVolume={Number(item.defaultvolume)}
+          defaultSpeed={Number(item.deafultspeed)}
```
bmartel, line 34:
```suggestion
          defaultSpeed={Number(item.defaultspeed)}
```

bmartel (changes requested):

src/tags/object/AudioNext/view.tsx
hlomzik: even more — `item.defaultzoom` and others will always be strings because of mst. and different places for default values are incosistent, and there is 3 different inconsistencies %)
- `defaultzoom` is 1 in mst, `WS_ZOOM_X.default` in the code; and there is also `WS_ZOOM_X.defaultValue` (no usages)
- `defaultspeed` is 1 in mst, 1 in the code, but there is `SPEED.default` constant
- `defaultvolume` is just 1 in both places

hlomzik (changes requested):
fix default values — we don't need them as they are set in mst; and use constants

hlomzik:
https://github.com/heartexlabs/label-studio-frontend/pull/818#discussion_r954799930

src/tags/object/AudioNext/view.tsx
```diff
@@ -14,9 +14,9 @@ const AudioNextView: FC<AudioNextProps> = ({ item }) => {
   const [position, setPosition] = useState(1);
   const [audioLength, setAudioLength] = useState(0);
 
-  const [zoom, setZoom] = useState(WS_ZOOM_X.default);
-  const [volume, setVolume] = useState(1);
-  const [speed, setSpeed] = useState(1);
+  const [zoom, setZoom] = useState(Number(item.defaultzoom ?? WS_ZOOM_X.default));
```
bmartel, line 17: I think what @hlomzik was referring to with default cases here is that it is redundant due to the definitions in Mobx State Tree already containing optional type and the same default. By the time the code gets to here, it already has the fallback handling taken care of.

Can reduce these checks to:

```suggestion
  const [zoom, setZoom] = useState(Number(item.defaultzoom));
```

bmartel (changes requested):

bmartel (approved):

hlomzik (approved):

PR #821, opened 2022-08-25
feat: DEV-2461: Add option to sync audio with paragraphs allowing playback of chunk position

bmartel:


yyassi-heartex (approved):
looks good to me, but should a FF be added for this as its appears to be a new feature.

src/tags/object/AudioNext/model.js
```diff
@@ -290,9 +290,9 @@ export const AudioModel = types.compose(
       },
 
       handleSeek() {
-        if (self._ws) {
-          self.triggerSyncSeek(self._ws.getCurrentTime());
-        }
+        if (!self._ws || self.syncedObject?.type === "paragraphs") return;
```
yyassi-heartex, line 293: need a tweak here to check for the FF - will need to import the FF higher up in the file too
```suggestion
        if (!self._ws || (isFF(FF_DEV_2461) && self.syncedObject?.type === "paragraphs")) return;
```

yyassi-heartex (changes requested):
minor tweak to check model.js for the FF aswell

e2e/tests/audio-paragraphs.test.js
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  I.click(`[aria-label="play-circle"]`);
+
+  I.click(`[aria-label="pause-circle"]`);
+
+
+  const [seekAudioPlusTime, seekParagraphAudioTime] = await AtAudioView.getCurrentAudioTime();
+
+  console.log({ hasFFDev2461, seekAudioPlusTime, seekParagraphAudioTime });
```
yyassi-heartex, line 130: please clean out the console.log

yyassi-heartex (approved):
approved, but please clean out the console.log - looks good otherwise

PR #822, opened 2022-08-25
fix: DEV-2918: Prevent annotating collapsed phrases in paragraphs

Gondragos:


yyassi-heartex (approved):
code looks good, one minor ask would be to do another scenario in the e2e test where the FF is set to false to make sure things run correctly (infact i think we should do that everytime we have a FF)

PR #839, opened 2022-09-08
feat: Add customizations to timeseries

ramberg:
Added optional markers to Start, End and Mid Points (circle, square, triangle-up and triangle-down)
Allowed functionality to disable the line (to switch between Scatter and Line+Scatter Plots)
Added the option for recognize NA / NULL in the value array (y-values) and remove that associated x-y-value from the data. This avoids an automatically inserted step that would "drag the line plot to zero, or add a marker" at that particular position.
Added the option to configure a custom range for both axis independent of the data ranges. This allows for fixed-scale axis for multiple tasks.
Added the option to hide the axis (incl. ticks and marks).
Added the width of the brush (selected portion of a plot), to the bottom right corner of the tracker (vertical cursor line).

### DIFF
--- a/src/tags/object/TimeSeries.js
+++ b/src/tags/object/TimeSeries.js
@@ -28,8 +28,6 @@ import PersistentStateMixin from "../../mixins/PersistentState";
 import "./TimeSeries/Channel";
 import { AnnotationMixin } from "../../mixins/AnnotationMixin";
 
-const moment = require("moment");
-
 /**
  * TimeSeries tag can be used to label time series data. Read more about Time Series Labeling on [the time series template page](../templates/time_series.html).
  *
@@ -66,6 +64,7 @@ const moment = require("moment");
  * @param {string} [timeColumn] Column name or index that provides temporal values. If your time series data has no temporal column then one is automatically generated.
  * @param {string} [timeFormat] Pattern used to parse values inside timeColumn, parsing is provided by d3, and follows `strftime` implementation
  * @param {string} [timeDisplayFormat] Format used to display temporal value. Can be a number or a date. If a temporal column is a date, use strftime to format it. If it's a number, use [d3 number](https://github.com/d3/d3-format#locale_format) formatting.
+ * @param {string} [durationDisplayFormat] Format used to display temporal duration value for brush range. If the temporal column is a date, use strftime to format it. If it's a number, use [d3 number](https://github.com/d3/d3-format#locale_format) formatting.
  * @param {string} [sep=,] Separator for your CSV file.
  * @param {string} [overviewChannels] Comma-separated list of channel names or indexes displayed in overview.
  * @param {string} [overviewWidth=25%] Default width of overview window in percents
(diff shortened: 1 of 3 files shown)


src/tags/object/TimeSeries.js
```diff
@@ -305,6 +307,17 @@ const Model = types
...
+    formatDuration(duration) {
+      if (!self._formatDuration) {
+        const { timedisplayformat: format, isDate } = self;
+
+        if (format === "date") self._formatDuration = duration => moment.utc(duration).format("HH:mm:ss.SSS");
+        else if (format) self._formatDuration = isDate ? d3.timeFormat(format) : d3.format(format);
+        else self._formatDuration = String;
+      }
+      return self._formatDuration(duration);
+    },
```
bmartel, line 320: I'm not sure what this function does, it appears to just hardcode a time format whenever the `timedisplayformat` is "date". 

Maybe instead of doing that, add an explicit `durationdisplayformat` and  default it to `%H:%M:%S.%L` so it's at least independently configurable and not tied to the timedisplayformat implicitly. 

If we were to do that, we could also skip the check for `format === "date"` and the moment dependency entirely, and just use `d3.utcFormat(format)`.

```suggestion
    formatDuration(duration) {
      if (!self._formatDuration) {
        const { durationdisplayformat: format } = self;

        if (format) self._formatDuration = d3.utcFormat(format);
        else self._formatDuration = String;
      }
      return self._formatDuration(duration);
    },
```

src/tags/object/TimeSeries.js
```diff
@@ -28,6 +28,8 @@ import PersistentStateMixin from "../../mixins/PersistentState";
 import "./TimeSeries/Channel";
 import { AnnotationMixin } from "../../mixins/AnnotationMixin";
 
+const moment = require("moment");
+
```
bmartel, line 32: Moment is not explicitly defined as a dependency of this project, and I don't think it is needed here.
```suggestion
```

bmartel (changes requested):
Seems good overall, some suggestions and observations.

src/tags/object/TimeSeries.js
hlomzik: that's the copy-paste from existing `formatTime` function couple of lines before, and yes, this function is weird %)
especially in this comparison with "date" which is not in the doc and not in the defaults, so no one will use it...

ramberg: I agree with the general concept to parameterize the format and get rid of the `format === "date"`, but don't we still need some type of distinction between non-date and date values (for ".1f" vs "%H:%M")? Then it is still somewhat coupled to timedisplayformat, which needs to be set to `date` (due to the functionality of `isDate`).

```suggestion
    formatDuration(duration) {
      if (!self._formatDuration) {
        const { durationdisplayformat: format, isDate } = self;

        if (format) self._formatDuration = isDate ? d3.utcFormat(format) : d3.format(format);
        else self._formatDuration = String;
      }
      return self._formatDuration(duration);
    },
```

src/tags/object/TimeSeries.js
ramberg: Agreed, will be removed by refactoring formatDuration.

src/tags/object/TimeSeries.js
bmartel: That makes sense for non time based ranges, yes. I think that suggestion should cover the cases required.

bmartel (approved):

bmartel:
closing in favour of #887

PR #847, opened 2022-09-13
fix: DEV-3405: Move zoom buttons to video's control panel

nicholasrq:


src/components/Timeline/Controls.tsx
```diff
@@ -53,6 +53,10 @@ export const Controls: FC<TimelineControlsProps> = memo(({
     return (position - 1) / frameRate;
   }, [position, frameRate]);
 
+  const customControls = useCustomControls(props.customControls);
+
+  console.log({ customControls });
```
yyassi-heartex, line 58: please remove console.log

src/components/Timeline/Types.ts
```diff
@@ -166,6 +168,11 @@ export interface TimelineControlsProps {
   onVolumeChange: TimelineProps["onVolumeChange"];
 }
 
+export interface TimelineCustomControls {
+  position: "left" | "right" | "left-center" | "right-center";
```
yyassi-heartex, line 172: do `"left" | "right" | "left-center" | "right-center"` map to anything? or get set from something else? would it be possible to change left-center and right-center to leftCenter and rightCenter respectively? this would make it so we can reference them above as `customControls?.rightCenter` instead of `customControls?.['right-center']`

src/tags/object/Video/HtxVideo.js
```diff
@@ -385,6 +369,28 @@ const HtxVideoView = ({ item }) => {
...
             controls={{ FramesControl: true }}
+            customControls={[
+              {
+                position: "left",
+                component: () => {
+                  return (
+                    <Dropdown.Trigger
+                      inline={isFullScreen}
+                      content={(
+                        <Menu size="medium" style={{ width: 150, minWidth: 0 }} closeDropdownOnItemClick={false}>
```
yyassi-heartex, line 380: anyway we can define a css rule to handle the width and min-width instead of inline styling it? these values are static

src/tags/object/Video/HtxVideo.js
```diff
@@ -385,6 +369,28 @@ const HtxVideoView = ({ item }) => {
...
+                          <Menu.Item onClick={zoomIn}>Zoom In</Menu.Item>
+                          <Menu.Item onClick={zoomOut}>Zoom Out</Menu.Item>
+                          <Menu.Item onClick={zoomToFit}>Zoom To Fit</Menu.Item>
+                          <Menu.Item onClick={zoomReset}>Zoom 100%</Menu.Item>
+                        </Menu>
+                      )}
+                    >
+                      <Button size="small" style={{ padding: 0 }}><IconZoomIn/></Button>
```
yyassi-heartex, line 388: like wise here for a css rule, if possible

yyassi-heartex (commented):
its very close - just a few relatively minor things

src/components/Timeline/Types.ts
nicholasrq: fair point, thank you

src/components/Timeline/Controls.tsx
```diff
@@ -53,6 +53,9 @@ export const Controls: FC<TimelineControlsProps> = memo(({
     return (position - 1) / frameRate;
   }, [position, frameRate]);
 
+  const customControls = useCustomControls(props.customControls);
+
+
```
yyassi-heartex, line 58:
```suggestion
```

yyassi-heartex: just need to clear out this extra space

yyassi-heartex (approved):
LGTM, just need to clear out the extra space there as I put in the suggestion

yyassi-heartex (approved):

PR #849, opened 2022-09-13
fix: DEV-3403: Allow frames scrubbing

nicholasrq:


src/components/Timeline/Views/Frames/Frames.tsx
```diff
@@ -170,6 +171,36 @@ export const Frames: FC<TimelineViewProps> = ({
     return value - currentOffsetX + timelineStartOffset;
   }, [position, currentOffsetX, step, length]);
 
+  const onFrameScrub = useCallback((e: MouseEvent) => {
+    const getMouseToFrame = (e: MouseEvent | globalThis.MouseEvent) => {
+      const mouseOffset = e.pageX - scrollable.current!.getBoundingClientRect().left - timelineStartOffset;
```
bmartel, line 176: `timelineStartOffset` should this not be in the useCallback deps array?

bmartel (commented):
Looks good, just one question

src/components/Timeline/Views/Frames/Frames.tsx
nicholasrq: no, it's not intended to change dynamically, so there's no need to update the callback

bmartel (approved):

PR #850, opened 2022-09-14
fix: DEV-3377: Scale image via size, not scale

hlomzik:
`transform: scale` leaves gaps on image border, so visually regions may look shifted.
Much better to change image sizes directly via width/height props.


hlomzik:
A little different issue, will be solved in #851

PR #852, opened 2022-09-14
fix: DEV-1943: After awhile, shapes disappear

yyassi-heartex:


src/components/Toolbar/Tool.js
```diff
@@ -70,14 +70,15 @@ export const Tool = ({
     currentShortcut = shortcut;
     if (shortcut && !hotkeys.hasKey(shortcut)) {
       hotkeys.addKey(shortcut, () => {
-        onClick?.();
+        if (!(tool?.annotation?.isDrawing ?? false))
```
bmartel, line 73:
```suggestion
        if (!(tool?.annotation?.isDrawing))
```

bmartel: its already boolean coerced, no need for the null coalesce.

src/components/Toolbar/Tool.js
```diff
@@ -106,17 +107,20 @@ export const Tool = ({
   }, [smart, extra]);
 
   const showControls = dynamic === false && controls?.length && (active || (controlsOnHover && hovered));
+  const isAnnotationDrawing = tool?.annotation?.isDrawing ?? false;
```
bmartel, line 110:
```suggestion
  const isAnnotationDrawing = tool?.annotation?.isDrawing;
```
same here

bmartel (changes requested):
Looks good, couple minor observations.

bmartel (approved):
Looks good, checked it and it works well 💪

hlomzik:
> /git merge develop
there is no `develop` in LSF :)

yyassi-heartex:
> > /git merge develop
> 
> there is no `develop` in LSF :)

lol yep as soon as i typed it im like "oh right" so i merged manually lol

src/tools/Base.js
```diff
@@ -24,13 +24,17 @@ const ToolView = observer(({ item }) => {
...
     smart: false,
+    isDrawingTool: false,
   })
   .volatile(() => ({
     dynamic: false,
     index: 1,
   }))
   .views(self => {
     return {
+      // get isDrawingTool() {
```
bmartel, line 35: I think this was meant to be removed

bmartel (approved):
Just the one commented out bit needs to get removed, approving this on that removal.

src/tools/Base.js
yyassi-heartex: you are right - i'll get rid of it

PR #855, opened 2022-09-16
fix: DEV-3429: Textarea placement in the config prevents OT

nicholasrq:


src/tags/object/Video/Video.js
```diff
@@ -158,7 +162,9 @@ const Model = types
     },
 
     addRegion(data) {
-      const control = self.control();
+      const control = self.videoControl() ?? self.control();
+
+      console.log({ control });
```
bmartel, line 167:
```suggestion
```

bmartel (changes requested):
Looks good, just the one leftover console

PR #858, opened 2022-09-16
fix: DEV-3428: Change to 1 item per page changing page by region

juliosgarbi:


bmartel (approved):

src/tags/object/PagedView.js
```diff
@@ -82,7 +82,8 @@ const HtxPagedView = observer(({ item }) => {
       if (obj.selected) {
         const _pageNumber = parseFloat(obj.object.name.split('_')[1]) + 1;
 
-        setPage(Math.ceil(_pageNumber / pageSize));
+        changePageSize(1);
```
bmartel, line 85: Is this needed if its always just 1 now?

src/tags/object/PagedView.js
```diff
@@ -82,7 +82,8 @@ const HtxPagedView = observer(({ item }) => {
...
+        setTimeout(() => setPage(_pageNumber));
```
bmartel, line 86: If you don't reset the pageSize state it wouldn't require a setTimeout here.

src/tags/object/PagedView.js
```diff
@@ -114,6 +115,11 @@ const HtxPagedView = observer(({ item }) => {
     };
   }, [page]);
 
+  const changePageSize = (pageSize) => {
```
bmartel, line 118: Not sure the solution here is to hardcode to 1, can you explain the performance issue with the scrolling to?

bmartel (changes requested):

src/tags/object/PagedView.js
juliosgarbi: In this case it is change the page size to 1 to show just the item that user choose!

src/common/Pagination/Pagination.tsx
```diff
@@ -13,6 +13,7 @@ interface PaginationProps {
   pageSize: number;
   totalPages: number;
   pageSizeOptions?: [];
+  paginated: boolean;
```
bmartel, line 16: this wording seems odd to parameterize the Pagination component with a `paginated` property. The definition of the component is that it is `paginated` always, but in this case we are changing the ability to select the `pageSize` not enabling/disabling the entire component. Maybe we change `paginated` to something like `pageSizeSelectable` as this is the only affected portion of the component.

bmartel (changes requested):

src/common/Pagination/Pagination.tsx
juliosgarbi: it makes sense!

bmartel (approved):

PR #861, opened 2022-09-20
fix: DEV-3436: Add comment icons in annotations dropdown

juliosgarbi:


yyassi-heartex (commented):
overall looking good, just a minor syntax fix

yyassi-heartex (approved):

src/components/TopBar/Annotations.js
```diff
@@ -48,6 +48,30 @@ export const Annotations = observer(({ store, annotationStore }) => {
     return () => document.removeEventListener('click', handleClick);
   }, []);
 
+  useEffect(() => {
```
bmartel, line 51: Are we still counting comments for each like this, I though the annotations now have the `unresolved_comment_count`, and `comment_count` attached when including those in the annotation query?

bmartel (changes requested):

src/components/TopBar/Annotations.js
juliosgarbi: it has unresolved_comment_count and comment_count attached but when we change something like adding new comment or resolving and unresolving a comment I need to refresh this information to change the icon in annotation dropdown, so I need to do this manually updating the informating using the commentStore

src/components/TopBar/Annotations.js
```diff
@@ -48,6 +48,30 @@ export const Annotations = observer(({ store, annotationStore }) => {
...
+
+    commentStore.comments.forEach(obj => {
+      _comments++;
+
+      if (!obj.isResolved) _unresolvedComments++;
+    });
+
+    commentStore.annotation.setUnresolvedCommentCount(_unresolvedComments);
+    commentStore.annotation.setCommentCount(_comments);
+  }, [JSON.stringify(commentStore)]);
```
nicholasrq, line 63: I would propose observing each individual property that needs to be tracked instead of running heavy `JSON.stringify`

you can use the `observe()` function imported from mobx for this purpose

it will be something like this:

```
useEffect(() => {
 const runOnPropertyChange = () => {
  /* whatever you want to do */
 }

 const disposers = [/* properties to observe */].map(prop => {
   return observe(commentStore, prop, runOnPropertyChange, true);
 })

 return () => {
  disposers.forEach(dispose => dispose());
 }
}, []);
```

another option is to try precomputing the number of resolved/unresolved comments in the memoized value and using it as a trigger for the `useEffect`:

```
const commentsNumber = useMemo(() => {
  let _unresolvedComments = 0;
  let _comments = 0;

  commentStore.comments.forEach(obj => {
    _comments++;

    if (!obj.isResolved) _unresolvedComments++;
  });

  return [_comments, _unresolvedComments];
}, [commentStore.comments]);

useEffect(() => {
  const [comments, unresolved] = commentsNumber;
  commentStore.annotation.setUnresolvedCommentCount(unresolved);
  commentStore.annotation.setCommentCount(comments);
}, [commentsNumber])
```

any other options instead of serializing the store are acceptable as well

nicholasrq (changes requested):

src/components/TopBar/Annotations.js
```diff
@@ -48,6 +48,36 @@ export const Annotations = observer(({ store, annotationStore }) => {
...
+
+    return [_comments, _unresolvedComments];
+  }, [[...commentStore.comments]]);
+
+  useEffect(() => {
+    const [comments, unresolved] = commentsNumber;
+
+    commentStore.annotation.setUnresolvedCommentCount(unresolved);
+    commentStore.annotation.setCommentCount(comments);
+  }, [commentsNumber]);
```
bmartel, line 69: This will rerun the memo every single time as it is just creating a new array each pass. We should probably look at another solution or remove the memo as this is will be no different than:

```suggestion
  const commentsNumber = (() => {
    let _unresolvedComments = 0;
    let _comments = 0;

    commentStore.comments.forEach(obj => {
      _comments++;

      if (!obj.isResolved) _unresolvedComments++;
    });

    return [_comments, _unresolvedComments];
  })();

  useEffect(() => {
    const [comments, unresolved] = commentsNumber;

    commentStore.annotation.setUnresolvedCommentCount(unresolved);
    commentStore.annotation.setCommentCount(comments);
  }, [commentsNumber]);
```

bmartel (changes requested):

bmartel (approved):

nicholasrq (approved):

PR #870, opened 2022-09-22
fix: DEV-3284: If video load times out display error message

Travis1282:
Added a new hook and reduced renders a bit, there are still a lot of extra renders
Clear timeout on every call to check loaded (there are always at least two) 
If video is not loaded in 3 seconds about load display error message and reset spinner

### DIFF
--- a/src/components/VideoCanvas/VideoCanvas.tsx
+++ b/src/components/VideoCanvas/VideoCanvas.tsx
@@ -390,7 +390,6 @@ export const VideoCanvas = memo(forwardRef<VideoRef, VideoProps>((props, ref) =>
     if (videoDimensions.ratio !== ratio) {
       const result = { ...videoDimensions, ratio };
 
-      console.log(result);
       setVideoDimensions(result);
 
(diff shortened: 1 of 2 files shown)


nicholasrq (approved):

src/components/VideoCanvas/VideoCanvas.tsx
```diff
@@ -398,9 +401,18 @@ export const VideoCanvas = memo(forwardRef<VideoRef, VideoProps>((props, ref) =>
 
   useEffect(() => {
     let isLoaded = false;
+    let timeout = undefined;
 
-    const checkVideoLoaded = () => {
+    const checkVideoLoaded: (counter: number) => NodeJS.Timeout | undefined = (counter) => {
       if (isLoaded) return;
+      if (counter > 300) {
```
yyassi-heartex, line 408: 300 is some what jarring here - would it be possible to create a constant variable so it describes what it is?

yyassi-heartex (commented):
very close, just a couple if very minor things

PR #872, opened 2022-09-23
feat: DEV-3346: Task history is stored on the backend for the label stream

juliosgarbi:


src/stores/AppStore.js
```diff
@@ -433,19 +433,30 @@ export default types
...
-          annotationId: null,
-        });
+        if (isPrevious) {
+          self.taskHistory = [
+            {
+              taskId: self.task.id,
+              annotationId: null,
+            },
+            ...self.taskHistory,
+          ];
```
bmartel, line 453: Is `taskHistory` a reactive object in mobx, or just a standard array? I would use the builtin array properties if its a mobx reactive value as I believe those are overridden in the case of mobx for reactive bindings propagation and replacing the array could mess up reactivity.

```suggestion
          self.taskHistory.unshift({
            taskId: self.task.id,
            annotationId: null,
          })
```

bmartel (changes requested):

bmartel (approved):

bmartel (approved):

hlomzik:
Implemented with a little different approach in https://github.com/HumanSignal/label-studio-frontend/pull/1181

PR #885, opened 2022-09-28
fix: DEV-3304: RUNTIME ERROR occurs when user tries to change value of 'defaultZoom', 'defaultSpeed' or 'defaultVolume' attributes for '<Audio/>' tag in 'Video Timeline Segmentation' template on Safari browser

yyassi-heartex:


bmartel (approved):

src/tags/object/AudioNext/view.tsx
```diff
@@ -13,9 +14,15 @@ const AudioNextView: FC<AudioNextProps> = ({ item }) => {
   const [position, setPosition] = useState(1);
   const [audioLength, setAudioLength] = useState(0);
 
-  const [zoom, setZoom] = useState(Number(item.defaultzoom));
-  const [volume, setVolume] = useState(Number(item.defaultvolume));
-  const [speed, setSpeed] = useState(Number(item.defaultspeed));
+  const numberify = (val: any, defaults: Record<string, number>) => {
```
bmartel, line 17: this can be lifted out of the component definition as it is a pure function.

yyassi-heartex: done

PR #889, opened 2022-10-03
fix: DEV-2993: Edit and Delete options are missing for Comments section

yyassi-heartex:


src/components/Comments/CommentFormBase.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const submitHandler = useCallback(async (e?: any) => {
+    e?.preventDefault?.();
+
+    if (!formRef.current) return;
+    
+    const comment = new FormData(formRef.current).get("comment") as string;
+    
+    if (!comment.trim()) return;
+
+    onSubmit?.(comment);
```
bmartel, line 41: Should probably just store the string as trimmed and pass that to submit.

```suggestion
    const comment = (new FormData(formRef.current).get("comment") as string)?.trim();
    
    if (!comment) return;

    onSubmit?.(comment);
```

src/components/Comments/CommentItem.tsx
```diff
@@ -9,13 +8,19 @@ import { Dropdown } from "../../common/Dropdown/Dropdown";
...
 import { Button } from "../../common/Button/Button";
 
 
 export const CommentItem: FC<{ comment: any }> = observer(({ comment }) => {
   const resolved = comment.isResolved;
+  const currentUser = window.APP_SETTINGS?.user;
+  const initialComment = comment.text;
+  const [currentComment, setCurrentComment] = useState(initialComment);
+
+  if (comment.isDeleted) return <></>;
```
bmartel, line 23: this should just return `null` here from what I can see, any reason for the empty fragment?

src/stores/Comment/Comment.js
```diff
@@ -39,7 +42,39 @@ export const Comment = types.model("Comment", {
...
+        text: comment,
+      });
+      self.setEditMode(false);
+    }
+
+    async function deleteComment() {
+      await self.sdk.invoke("comments:delete", {
+        id: self.id,
+      });
+      self.setDeleted(true);
```
bmartel, line 69: Probably should reset the confirmation as well here.

```suggestion
      self.setDeleted(true);
      self.setConfirmMode(false);
```

bmartel (changes requested):

bmartel (approved):

PR #890, opened 2022-10-03
fix: DEV-3551: Fix bug when zoom on large videos corrupts regions

nicholasrq:


src/components/VideoCanvas/VideoCanvas.tsx
```diff
@@ -45,7 +45,7 @@ type VideoDimentions = {
   ratio: number,
 }
 
-const clampZoom = (value: number) => clamp(value, 0.25, 16);
+export const clampZoom = (value: number) => clamp(value, 0.1, 16);
```
yyassi-heartex, line 48: Can we get a constant variable to indicate what the 0.1 and 16 means - these numbers just appears some what arbitrary

src/tags/object/Video/HtxVideo.js
```diff
@@ -175,11 +170,11 @@ const HtxVideoView = ({ item }) => {
   }, [panMode, pan]);
 
   const zoomIn = useCallback(() => {
-    setZoom(clamp(zoom + 0.1, 0.25, 16));
+    setZoom(zoom + 0.1);
```
yyassi-heartex, line 173: since 0.1 is used in zoomIn and zoomOut it would be great to create a constant variable called something like `interval` or `zoomInterval` that we can reuse

yyassi-heartex (commented):
looking good - just a couple of very minor things to improve readability

yyassi-heartex (approved):

PR #892, opened 2022-10-04
feat: DEV-1041: Configurable hop size in VOT

nicholasrq:
Add Video hop setting


src/components/Settings/Settings.js
```diff
@@ -71,9 +70,84 @@ const TabPaneGeneral = (store) => {
...
+      {/* <br/> */}
+      {/* <Checkbox */}
+      {/*   value="Show image in fullsize" */}
+      {/*   defaultChecked={store.settings.imageFullSize} */}
+      {/*   onChange={() => { */}
+      {/*     store.settings.toggleImageFS(); */}
+      {/*   }} */}
+      {/* > */}
+      {/*   Show image in fullsize */}
+      {/* </Checkbox> */}
```
yyassi-heartex, line 123: is the intent of the commented out block here to put it in at some point later? or is it meant to be gone? if not needed it maybe best to remove it

src/components/Settings/Settings.js
```diff
@@ -71,9 +70,84 @@ const TabPaneGeneral = (store) => {
...
+        checked={store.settings.bottomSidePanel}
+        onChange={() => {
+          store.settings.toggleBottomSP();
+          setTimeout(triggerResizeEvent);
+        }}
+      >
+            Move sidepanel to the bottom
+      </Checkbox>
+
+      <br />
```
yyassi-heartex, line 88: rather than use `<br />`s can we wrap the checkboxes in `<Elem>` and potentially apply css as needed? maybe with a `<Block>` above instead of the `<>` so we can be more specific

src/tags/object/Video/HtxVideo.js
```diff
@@ -22,8 +22,9 @@ import { IconZoomIn } from "../../../assets/icons";
 
 // const hotkeys = Hotkey("Video", "Video Annotation");
```
yyassi-heartex, line 23: please remove this commented out line

yyassi-heartex (commented):
looking good overall just a few minor nit picky stuff for cleanup

src/components/Settings/Settings.js
nicholasrq: I'm not sure what this thing is and why it's commented out. I just moved this piece of code from the main Settings component. Decided to keep it that way to avoid any bugs.

src/components/Settings/Settings.js
nicholasrq: Fair point. I didn't want to touch this part as it's out of scope of the task, but can definitely do that.

yyassi-heartex (approved):

PR #899, opened 2022-10-06
fix: DEV-3341: selected=true <Choice> attribute is not supported for Taxonomy

yyassi-heartex:


bmartel (approved):

e2e/tests/taxonomy.test.js
```diff
@@ -0,0 +1,102 @@
...
+      </Taxonomy>
+    </View>`,
+    text: `To have faith is to trust yourself to the water`,
+    annotations: [
+      { label: 'PER', rangeStart: 0, rangeEnd: 2, text: 'To', taxonmy: ["Eukarya", "Extraterrestial"], taxonomyLabel: "Eukarya / Extraterrestial" },
+    ],
+  },
+};
+
+const taxonomyTable = new DataTable(["taxnomyName"]);
```
hlomzik, line 61: please, fix all the typos "taxnomy" and "taxonmy"

e2e/tests/taxonomy.test.js
```diff
@@ -0,0 +1,102 @@
...
+      selector: ".lsf-htx-richtext",
+      rangeStart: annotation.rangeStart,
+      rangeEnd: annotation.rangeEnd,
+    });
+    I.click(locate("span").withText(annotation.label));
+    I.click(locate("li").withText(annotation.text));
+  });
+
+  annotations.forEach(annotation => {
+    I.click(locate("li").withText(annotation.text));
```
hlomzik, line 88: duplication from previous block, but there are questions anyway:
- why do we need to select every region without checking anything? at the end there will be one only region.
- this would not work with new outliner enabled

e2e/tests/taxonomy.test.js
```diff
@@ -0,0 +1,102 @@
...
+        <Choice value="Eukarya">
+          <Choice value="Human" />
+          <Choice value="Oppossum" />
+          <Choice value="Extraterrestial" selected="true" />
+        </Choice>
+      </Taxonomy>
+    </View>`,
+    text: `To have faith is to trust yourself to the water`,
+    annotations: [
+      { label: 'PER', rangeStart: 0, rangeEnd: 2, text: 'To', taxonmy: ["Eukarya", "Extraterrestial"], taxonomyLabel: "Eukarya / Extraterrestial" },
```
hlomzik, line 56: for both cases this preselected choice is also presented in annotation, so it doesn't check preselection. there are should be tests with preselected choice and 3 different cases: no annotation at all (preselected shown), annotations without this choice (should not be shown), annotation with this choice (should be shown, just to check)

hlomzik (changes requested):

src/tags/object/RichText/view.js
```diff
@@ -83,7 +83,11 @@ class RichTextPieceView extends Component {
       normedRange.text = selectionText;
       normedRange.isText = item.type === "text";
       normedRange.dynamic = this.props.store.autoAnnotation;
-      item.addRegion(normedRange, this.doubleClickSelection);
+
+      const selectedAnnotation = window.Htx.annotationStore.selected.toNames.get(item.name).toJSON()?.filter(annotation => annotation.perregion && (annotation.preselectedValues?.length ?? 0) > 0);
```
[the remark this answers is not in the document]
yyassi-heartex: I think that would be fine, @hlomzik what are your thoughts on that as I used `Htx. annotationStore` after your suggestion but I believe technically both options are fine since both point to the same object. comes down to what direction we want to reproduce in the code going forward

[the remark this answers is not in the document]
hlomzik: Yep, `Htx` is only for external usage. The fact that we use it in Area and Image is just a dirty hack I'd like to get rid of. It should have a better comment about it =(
Also using `selected` to get connected annotation is also a bad practice, I'm fixing this right now.
So just do the same as in other places of this file — `item.annotation`

e2e/tests/taxonomy.test.js
```diff
@@ -0,0 +1,293 @@
...
+      } },
+    ],
+    isPerRegion: true,
+    FF: {
+      ff_front_1170_outliner_030222_short: true,
+    },
+  },
+};
+
+const taxonomyTable = new DataTable(["taxnomyName"]);
```
hlomzik, line 217: typo "taxnomy"

hlomzik (changes requested):

bmartel (approved):

hlomzik:
fixed in https://github.com/HumanSignal/label-studio-frontend/pull/1239 with tests for cypress

PR #900, opened 2022-10-06
fix: DEV-3284: Await the wait for all objects to be ready before testing video regions

bmartel:


bmartel:
No longer required, the test is fixed by #901

PR #902, opened 2022-10-07
fix: DEV-3293: Enable alias for taxonomy choice

Travis1282:
enable alias for taxonomy, use label for displayed path when internal path is alias

### DIFF
--- a/src/components/Taxonomy/Taxonomy.tsx
+++ b/src/components/Taxonomy/Taxonomy.tsx
@@ -4,7 +4,7 @@ import { Dropdown, Menu } from "antd";
 import { useToggle } from "../../hooks/useToggle";
 import { isArraysEqual } from "../../utils/utilities";
 import { LsChevron } from "../../assets/icons";
-import TreeStructure, { RowItem } from "../TreeStructure/TreeStructure";
+import TreeStructure from "../TreeStructure/TreeStructure";
 
 import styles from "./Taxonomy.module.scss";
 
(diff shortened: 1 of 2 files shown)


src/tags/control/Taxonomy.js
```diff
@@ -83,7 +83,7 @@ function uniq(nodes) {
 function traverse(root) {
   const visitNode = function(node, parents = []) {
     const label = node.value;
-    const path = [...parents, label]; // @todo node.alias || label; problems with showFullPath
+    const path = [...parents, node.alias ?? node.value]; // @todo node.alias || label; problems with showFullPath
```
bmartel, line 86: this comment can be removed now that it's implemented?
```suggestion
    const path = [...parents, node.alias ?? label];
```

src/components/Taxonomy/Taxonomy.tsx
```diff
@@ -113,13 +113,22 @@ const UserLabelForm = ({ onAddLabel, onFinish, path }: UserLabelFormProps) => {
...
+        const label = flatItems.find(taxonomyItem => taxonomyItem.path[taxonomyItem.path.length - 1] === value)?.label;
+        
+        return label ?? value;  
+      }),
+  );
+  
   return (
     <div className={styles.taxonomy__selected}>
-      {selected.map(path => (
+      {selectedLabels.map(path => (
```
bmartel, line 131: I'm curious what this does when @yyassi-heartex 's changes are in the mix? I'm thinking this should be tested in conjunction as well.

bmartel (changes requested):

src/components/Taxonomy/Taxonomy.tsx
Travis1282: me too but someone has to merge first

yyassi-heartex: agreed, if this merges first i can resolve any conflicts that come up in my branch - no biggie

Travis1282: Brandon, that was a good catch about the chip deletion. I have updated to handle this too. ready to merge

yyassi-heartex (approved):

PR #903, opened 2022-10-07
fix: DEV-3559: Taxonomy annotations are not displayed after selection

juliosgarbi:


hlomzik (approved):

hlomzik:
It's better to just unselect items which are not in the list anymore (when you created region and them undo this), but it's for later, now it's also good, but less errorproof though.

PR #910, opened 2022-10-12
fix: DEV-3594: Check has selected objects 

Travis1282:


bmartel (approved):

hlomzik:
Had some issues and the original issue was solved by #915

PR #920, opened 2022-10-21
fix: DEV-3564: Annotations drawn for audio region should span all channels

bmartel:


src/regions/AudioRegion.js
```diff
@@ -136,6 +141,10 @@ const Model = types
       }
 
       el.className = classes.filter(Boolean).join(" ");
+
+      // Annotation must visually be drawn across both channels
+      el.style.top = "0px";
+      el.style.height = "100%";
```
yyassi-heartex, line 147: any way we can add these styles through the css instead of inline?

yyassi-heartex (approved):

PR #928, opened 2022-10-25
feat: DEV-3391: Render real Annotations in View All

hlomzik:
- [ ] reduce traverses in `Tree#extractNames()`
- [ ] check Channel without ff
- [ ] add comment about these ideas/concept
- [ ] fix annotation history selected
- [ ] fix Tools? (using the same hack through store will help, but will throw us back to the original problem)
- [ ] mysterious region selected in prediction in View All

PR #933, opened 2022-10-27
fix: DEV-3644: Readonly regions should always be selectable

bmartel:


Travis1282 (approved):

hlomzik (changes requested):
~~Overall is great, but Taxonomy is editable in readonly~~
Oh nevermind, in readonly Taxonomy is really readonly because dropdown is not rendered at all... And now I understand that we have confusion between `readonly` and `editable`, so we need to discuss it.

hlomzik:
Merged branch of DEV-3391 and DEV-3644: https://github.com/heartexlabs/label-studio-frontend/tree/fb-dev-3391/readonly

bmartel:
@hlomzik 
> ~Overall is great, but Taxonomy is editable in readonly~ Oh nevermind, in readonly Taxonomy is really readonly because dropdown is not rendered at all... And now I understand that we have confusion between `readonly` and `editable`, so we need to discuss it.

This behaviour is what resulted in honouring the readonly nature of the annotation from taxonomy perspective. If this is not correct we should discuss what is expected of the related components. (this likely is an issue with more of the inner components of various configs).

bmartel:
What do we want to do with this branch @hlomzik? These commits are in #928, but if we want to cherry pick the readonly changes it won't be possible without a squash commit of just these changes.

PR #935, opened 2022-10-27
fix: DEV-3617: Memory leaks in Taxonomy + Repeater cases

nicholasrq:
This PR provides two fixes for the memory-intensive cases:

- large taxonomy with in-config choices
- large taxonomy within repeater with dynamic children

for the first case the fix will work automatically

for the second case a special attribute was introduced called `sharedStore`. it accepts a string identifier:

```xml
<View>
  <Header size="10">Document ID: $document_id</Header>
  <Repeater indexFlag="{{idx}}" on="$images" mode="pagination">
    <View style="display: flex; margin-bottom: 10px;">
      <Image maxHeight="auto" name="page_{{idx}}" value="$images[{{idx}}].url"/>

      <RectangleLabels name="taxonomy_{{idx}}" toName="page_{{idx}}">
        <Label alias="taxonomy" background="#B2CDFE" selectedColor="#74A7FE" value="Primary taxonomy">
        </Label>
        <Label alias="taxonomy_other" background="#F9AEC8" selectedColor="#FF7FAC" value="Other taxonomies">
        </Label>
      </RectangleLabels>

      <Taxonomy name="categories_{{idx}}" perRegion="true" toName="page_{{idx}}" visibleWhen="region-selected" whenLabelValue="taxonomy" value="$categories" sharedStore="categories"/>
      <Taxonomy name="taxonomy_other_{{idx}}" perRegion="true" toName="page_{{idx}}" visibleWhen="region-selected" whenLabelValue="taxonomy_other" value="$other" sharedStore="taxonomy_other"/>
    </View>
  </Repeater>
</View>
```


src/mixins/SharedChoiceStore/extender.js
```diff
@@ -0,0 +1,16 @@
...
+export const StoreExtender = types.model("StoreExtender", {
+  sharedStores: types.optional(types.map(SharedStoreModel), {}),
+}).actions((self) => ({
+  addSharedStore(store) {
+    self.sharedStores.set(store.id, store);
+  },
+  beforeDestroy() {
+    self.sharedStores.forEach((store) => {
+      detach(store);
+      store.die();
```
bmartel, line 13: What a dramatic method 😆

src/mixins/SharedStoreMixin.js
```diff
@@ -0,0 +1,63 @@
+import { detach, types } from "mobx-state-tree";
+import Types from "../core/Types";
+
+const SharedStore = {};
+
+const SharedStoreModel = (childrenType) => types.model({
```
bmartel, line 6: What is the correlation between these set of SharedStore primitives and the set within SharedChoiceStore? Seems like these are basically the same, do we need both?

src/stores/RegionStore.js
```diff
@@ -6,6 +6,7 @@ import { AllRegionsType } from "../regions";
 import { debounce } from "../utils/debounce";
 import Tree, { TRAVERSE_STOP } from "../core/Tree";
 import { FF_DEV_2755, isFF } from "../utils/feature-flags";
+import { toJS } from "mobx";
```
bmartel, line 9: Seems like this is unused?

src/mixins/SharedStoreMixin.js
bmartel: Also, the SharedChoiceStore seems very specific to include directly into AnnotationStore, is this the intent? Seeing the factory based (and agnostic) SharedStore seems more flexible?

nicholasrq: so, `SharedStore` is more like a cache because we're creating a store within the `preProecssSnapshot` where there's no access to the tree. after that, when the node is created, the store is retrieved from the cache and attached to the tree

speaking of the AnnotationStore, it's done on purpose. attaching the store to the taxonomy itself will not solve the problem because every page has its own set of tags. the whole concept of shared storage doesn't work in this case

i had an idea of attaching it to the annotation, but it doesn't make sense as well – the list of choices can only change on the task level. keeping shared stores on the AnnotationStore level allows skipping the loading phase on other annotations

unfortunately, there's no other place between annotation and annotation store to fit the stores in

bmartel: Ok so we are keeping the concept of SharedChoiceStore, and that makes sense if its a global concept to have in the AnnotationStore. The one follow up to that which I was wondering, is where is this `[src/mixins/SharedStoreMixin.js](https://github.com/heartexlabs/label-studio-frontend/pull/935/files/7b441225e3466688e0e206f1c162021efeb573e4#diff-ea82889da1ec3e6f589dc270ac9d2aecf064ccf35b5cf6a298d8e464799a7dd9)` used? All I see that we are using is the [src/mixins/SharedChoiceStore/extender.js](https://github.com/heartexlabs/label-studio-frontend/pull/935/files/7b441225e3466688e0e206f1c162021efeb573e4#diff-bcd52178dea55c61de0ac21e27865af21f94f1980b98ed19723dc34df82ebe88)

nicholasrq: it's a mixin that is composed inside the [`AnnotationStore`](https://github.com/heartexlabs/label-studio-frontend/pull/935/files/7b441225e3466688e0e206f1c162021efeb573e4#diff-2aba458b602dc32bb13ce2d9e19b257d79a23f6eaab9e3d9903814b809066cd4R482-R485)

bmartel: That one I see as the canonical, but this mixin we are commenting in, its not found or used anywhere I can see? I searched for imports for `mixins/SharedStoreMixin` and I don't see this used at all.

src/mixins/DynamicChildrenMixin.js
```diff
@@ -26,29 +26,45 @@ const DynamicChildrenMixin = types.model({
     };
 
     return {
+      updateWithDynamicChildren(data) {
+        const list = (self.children ?? []).concat(prepareDynamicChildrenData(data));
+
+        self.children = list;
```
bmartel, line 32: This setter in the shared store will lock, but I don't see a corresponding unlock in this mixin. What controls the unlock of the shared store here?

nicholasrq: it's intended. the store is unlocked by default, and it's not supposed to be unlocked after children are set. otherwise, other taxonomies will do the same work over and over again

bmartel (approved):

PR #937, opened 2022-10-28
fix: DEV-1598: Fix for empty toname in control tag

juliosgarbi:


src/stores/Annotation/Annotation.js
```diff
@@ -746,7 +756,7 @@ export const Annotation = types
         results: [result],
       };
 
-      const area = self.areas.put(areaRaw);
+      const area = self?.areas?.put(areaRaw);
```
bmartel, line 759: This change makes the below change no longer safer. `area` can now be undefined, so we will need to make sure the below checks `!area.classification` uses `area && !area.classification`.

bmartel: Similarily the line below it, maybe just early return `if (!area) return`.
```suggestion
      const area = self?.areas?.put(areaRaw);
      if (!area) return;
```

bmartel (changes requested):

bmartel (approved):

src/stores/Annotation/Annotation.js
hlomzik: Why do we need this safe reference? `self` is definitely defined and `self.areas` is defined mst required property.

hlomzik (changes requested):
Small fix for ?. safe reference

src/stores/Annotation/Annotation.js
hlomzik: discussed it — weird mst bug possibly because of production build, so leave it as is

hlomzik (approved):

PR #941, opened 2022-10-31
Feat: DEV-3643: Region click repeater navigation tests

Travis1282:


bmartel (changes requested):

e2e/tests/repeater-paginate.test.js
```diff
@@ -0,0 +1,212 @@
...
+        origin: "manual",
+        to_name: "page_1",
+        from_name: "labels_1",
+        image_rotation: 0,
+        original_width: 600,
+        original_height: 776,
+      },
+    ],
+  },
+];
```
bmartel, line 212: This one is still out of order

bmartel (changes requested):

e2e/tests/repeater-paginate.test.js
bmartel: Moved that one as well. Curious how it works with the data being out of order of declaration and usage. Typescript and the browser aren't able to continue past seeing that type of thing. Unless webpack's compilation and hoisting is fixing it at some point.

e2e/tests/repeater-paginate-scroll.test.js
```diff
@@ -0,0 +1,217 @@
...
+});
+
+Scenario("Regions will scroll view window on region click", async function({ I, LabelStudio }) {
+  const params = { config: configScroll, annotations, data };
+  
+  LabelStudio.setFeatureFlags({
+    ff_front_1170_outliner_030222_short: true,
+  });
+  I.refreshPage();
+  I.amOnPage("/");
```
Gondragos, line 204:
```suggestion
  I.refreshPage();
  I.amOnPage("/");
  LabelStudio.setFeatureFlags({
    ff_front_1170_outliner_030222_short: false,
  });

```
It does not change anything but it would be easier to interpretate.

bmartel (approved):

e2e/tests/repeater-paginate-scroll.test.js
Travis1282: for some reason this does not pass

e2e/tests/repeater-paginate-scroll.test.js
```diff
@@ -0,0 +1,216 @@
...
+  I.waitForVisible(locate(".lsf-label__hotkey").withText("2"));
+
+});
+
+Scenario("Regions will scroll view window on region click", async function({ I, LabelStudio }) {
+  const params = { config: configScroll, annotations, data };
+
+  I.amOnPage("/");
+  LabelStudio.setFeatureFlags({
+    ff_front_1170_outliner_030222_short: true,
```
Gondragos, line 202:
```suggestion
    ff_front_1170_outliner_030222_short: false,

```
I think right now it fails because of that.

Travis1282: face palm. yes. yes, you are right

PR #953, opened 2022-11-04
fix: DEV-3688: Start select current area when create or resume drawing

juliosgarbi:


src/components/ImageView/ImageView.js
```diff
@@ -500,12 +500,12 @@ export default observer(
         this.handleDeferredMouseDown = handleDeferredMouseDown;
         this.handleDeferredMouseDown?.();
         this.handleDeferredMouseDown = null;
-      }, 100);
+      }, this.props.item.annotation.isDrawing ? 0 : 100);
```
bmartel, line 503: This seems interesting, I am gonna try this locally.

PR #954, opened 2022-11-07
fix: DEV-3692: Comment edited session state

Travis1282:
If edited display edited time in session

### DIFF
--- a/src/components/Comments/CommentItem.tsx
+++ b/src/components/Comments/CommentItem.tsx
@@ -1,5 +1,5 @@
 import { observer } from "mobx-react";
-import { FC, useEffect, useState } from "react";
+import { FC, useState } from "react";
 import { Tooltip } from "antd";
 import { IconCheck, IconEllipsis } from "../../assets/icons";
 import { Space } from "../../common/Space/Space";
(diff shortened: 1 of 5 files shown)


src/components/Comments/CommentItem.tsx
```diff
@@ -19,9 +19,27 @@ export const CommentItem: FC<{ comment: any }> = observer(({ comment }) => {
...
+    else if (isEdited === false) setIsEdited((new Date()));
+  }, [comment.isEditMode]);
 
+  const TimeTracker = () => {
+    const time = isEdited ? isEdited : comment.createdAt;
+  
+    return comment.isPersisted && comment.createdAt && (
+      <Elem name="date">
+        <Tooltip placement="topRight" title={new Date(time).toLocaleString()}>
+          {`${isEdited ? "edited" : "created"} ${humanDateDiff(time)}`}
```
bmartel, line 37: just a nitpick, but it seems redundant to put the word `created` for every comment. The specific clarification is the edited state, so I would think this would be better to just add `edited` when its the differing state.

bmartel (changes requested):

src/components/Comments/CommentItem.tsx
```diff
@@ -19,9 +19,27 @@ export const CommentItem: FC<{ comment: any }> = observer(({ comment }) => {
   const currentUser = window.APP_SETTINGS?.user;
   const initialComment = comment.text;
   const [currentComment, setCurrentComment] = useState(initialComment);
+  const [isEdited, setIsEdited] = useState<Date | boolean | undefined>();
 
   if (comment.isDeleted) return null;
+  
+  useEffect(() => {
+    if (comment.isEditMode) setIsEdited(false);
+    else if (isEdited === false) setIsEdited((new Date()));
```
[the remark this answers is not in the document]
Travis1282: there is no updated at timestamp unfortunately

src/components/Comments/Comments.tsx
```diff
@@ -46,7 +46,7 @@ export const Comments: FC<{ commentStore: any, cacheKey?: string }>= observer(({
   return (
     <Block name="comments">
       <CommentForm commentStore={commentStore} inline />
-      <CommentsList commentStore={commentStore} />
+      <CommentsList commentStore={commentStore} loadComments={loadComments} />
```
bmartel, line 49: `loadComments` is never used and can be removed

src/components/Comments/CommentItem.tsx
```diff
@@ -25,18 +24,20 @@ interface Comment {
     createdBy: any,
     text: string,
     isResolved: boolean,
+    loadComments: () => void,
```
bmartel, line 27: comment model doesn't have `loadComments`

src/stores/Comment/CommentStore.js
```diff
@@ -271,8 +271,7 @@ export const CommentStore = types
     }
 
     const listComments = flow((function* ({ mounted = { current: true } } = {}) {
-      self.setComments([]);
-
+      if (!mounted.current) self.setComments([]);
```
bmartel, line 274: I'm a bit iffy on reusing the mounted state like this, as its supposed to be a safeguard against setting data on unmounted components, that could otherwise be attempting to set data on a now removed part of the state tree. What was the use case of this change?

bmartel (changes requested):

src/stores/Comment/CommentStore.js
Travis1282: The comments would blink off when making the list call after edit. I can maybe pass another optional value here

bmartel (approved):

PR #1002, opened 2022-11-09
fix: DEV-3693: Explore reviews module show buggs

yyassi-heartex:


e2e/tests/unfinished-polygons.test.js
```diff
@@ -103,9 +103,12 @@ Scenario("Saving polygon drawing steps to history", async function({ I, LabelStu
...
   I.say(`try to undo closing and 2 last points`);
-  I.pressKey(['CommandOrControl', 'Z']);
-  I.pressKey(['CommandOrControl', 'Z']);
-  I.pressKey(['CommandOrControl', 'Z']);
+  I.click("button[aria-label=Undo]");
+  I.click("button[aria-label=Undo]");
+  I.click("button[aria-label=Undo]");
+  // I.pressKey(['CommandOrControl', 'Z']);
+  // I.pressKey(['CommandOrControl', 'Z']);
+  // I.pressKey(['CommandOrControl', 'Z']);
```
bmartel, line 111: Are these meant to stay or be replaced strictly by the button usage?

Maybe can we instead split to another test to ensure both work correctly in the same fashion?

bmartel (changes requested):
Looks good, just one suggestion with the unfinished polygon's test.

e2e/tests/unfinished-polygons.test.js
yyassi-heartex: they should be removed, as currently the hotkey undo is not working the same way as clicking the undo button and it should be addressed in another ticket

bmartel (approved):

bmartel (approved):

bmartel (approved):

PR #1005, opened 2022-11-10
fix: DEV-1852: Fix missing labels on video regions

Gondragos:


src/tags/object/Video/Rectangle.js
```diff
@@ -21,46 +22,56 @@ const RectanglePure = ({ reg, frame, workingArea, ...rest }) => {
...
+        stroke={style.strokeColor}
+        strokeScaleEnabled={false}
+        opacity={reg.hidden ? 0 : 1}
+        onTransformEnd={e => {
+          const node = e.target;
+          const scaleX = node.scaleX();
+          const scaleY = node.scaleY();
+          // set minimal value
+          const w = Math.max(3, node.width() * scaleX);
+          const h = Math.max(3, node.height() * scaleY);
```
yyassi-heartex, line 58: can we use a const variable to define what 3 is - the 3 is just in there and I dont understand what it represents (aside from being a safe value in case node dimensions are too small)

yyassi-heartex (commented):
its looking good over all, just one minor request to use a constant value

src/tags/object/Video/Rectangle.js
Gondragos: Sure! I think I have found the right constant. And yeah, it may be too small, but it was done not by me and it is the same for the ImageView regions. So I think for now it is the right value.

yyassi-heartex (approved):

PR #1012, opened 2022-11-10
fix: DEV-3672: I can't reassign label to bounding box with a hotkey

yyassi-heartex:


src/tags/control/Label.js
```diff
@@ -236,12 +236,18 @@ const Model = types.model({
...
+    return self.onLabelInteract();
+  },
+
+  onClick() {
+    return self.onLabelInteract();
+  },
+
+  onLabelInteract() {
+    if(!self.annotation.isDrawing) {
+      self.toggleSelected();
```
bmartel, line 248: what is the return of `toggleSelected`? I seen this was previously returning the result of that function in `onHotKey`. Just want to make sure this is returning the equivalence as previous in all scenarios.

yyassi-heartex: appears to return nothing based on what I see here https://github.com/heartexlabs/label-studio-frontend/blob/0aafae54f03808dddbce54642c1aac281ea487d4/src/tags/control/Label.js#L114

yyassi-heartex: I opted to return false as the onClick was returning false

hlomzik (approved):

PR #1016, opened 2022-11-15
fix: DEV-3603: Fix serialization calculation for regions so that values remain an exact match when compared

bmartel:


hlomzik:
Got obsolete after https://github.com/HumanSignal/label-studio-frontend/pull/1148, these methods will be removed soon

PR #1018, opened 2022-11-15
fix: DEV-3733: Persist and use the pending draft to attach added comments

bmartel:


src/stores/Comment/CommentStore.js
```diff
@@ -111,6 +111,7 @@ export const CommentStore = types
     }
 
     function removeCommentById(id)  {
+
```
yyassi-heartex, line 114: Just a little cleanup here
```suggestion
```

yyassi-heartex (approved):
looks good to me, just optional minor clean up

PR #1020, opened 2022-11-17
fix: DEV-3445: Regions and segments not resizable if at start

yyassi-heartex:


src/lib/AudioUltra/Regions/Segment.ts
```diff
@@ -230,24 +230,27 @@ export class Segment extends Events<SegmentEvents> {
   };
 
   private handleDrag = (e: MouseEvent) => {
-    if (!this.updateable) return;
```
bmartel, line 233: was this meant to get deleted? This is needed for blocking updates to regions.

bmartel (changes requested):

bmartel (approved):

PR #1022, opened 2022-11-18
fix: DEV-2706: Fix the misalignment between frame numbers

nicholasrq:


src/components/Timeline/SideControls/FramesControl.tsx
```diff
@@ -14,21 +14,23 @@ export const FramesControl: FC<TimelineSideControlProps> = ({
     return length - 1;
   }, [length]);
 
+  console.log({ duration, length });
```
juliosgarbi: Leftover console

juliosgarbi (approved):
LGTM

PR #1023, opened 2022-11-18
feat: DEV-3638: Numpad hotkeys enabled

Travis1282:
dispatch synthetic keyboard event for keyboard number corresponding to numpad number


src/core/Hotkey.ts
```diff
@@ -40,11 +40,15 @@ const _hotkeys_desc: { [key: string]: string } = {};
 const _namespaces: {[key: string]: HotkeyNamespace} = {};
 const _destructors: (() => void)[] = [];
 
-keymaster.filter = function(event) {
+const translateNumpad = (event:any) => 
+  document.dispatchEvent(new KeyboardEvent('keydown', { keyCode:  event.keyCode - 48 }));
```
bmartel, line 44: Can we use keycode constants instead of just the numbers so it is easier to see these ranges and what we are translating?

src/core/Hotkey.ts
```diff
@@ -40,11 +40,15 @@ const _hotkeys_desc: { [key: string]: string } = {};
...
+
+keymaster.filter = function(event: any) {
   if (keymaster.getScope() === "__none__") return false;
 
   const tag = (event.target || event.srcElement)?.tagName;
 
+  if (event.keyCode >= 96 && event.keyCode <= 105) translateNumpad(event);
```
bmartel, line 51: here as well

bmartel (approved):

src/core/Hotkey.ts
Travis1282: I changed to me more semantic and easier to read the keyCodes are what is used by keymaster.js so keycode constants won't work

PR #1027, opened 2022-11-22
merging ls-waveform changes

bmartel:
This should also fix: 
- https://heartex.atlassian.net/browse/DEV-3781
- enable auto playback of created segments
- showing selected regions along the timeline
- addRegion api to correctly render and return the added segment or region to the waveform caller


bmartel:
closing in favor of #1029

PR #1035, opened 2022-11-28
fix: DEV-3790: all strings color from Label is working

juliosgarbi:


bmartel (approved):

bmartel (changes requested):
I just noticed the original target was master which made this incorrectly build against develop for both LSE/LSO.

PR #1038, opened 2022-11-29
fix: DEV-3748: Fix annotation disappear on unto

Travis1282:
block empty recorder states

### DIFF
--- a/src/core/TimeTraveller.js
+++ b/src/core/TimeTraveller.js
@@ -94,6 +94,7 @@ const TimeTraveller = types
 
           return;
         }
+        if (Object.keys(recorder).length < 1) return;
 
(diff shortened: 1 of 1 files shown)


src/core/TimeTraveller.js
```diff
@@ -102,6 +103,7 @@ const TimeTraveller = types
         changesDuringFreeze = false;
       },
 
+
```
bmartel, line 106:
```suggestion
```

bmartel (approved):
Just the space, this should be good I would just watch the tests to make sure some other code hasn't taken this into account since this seems like its existed forever.

bmartel (changes requested):
branch name and title should be based on DEV-2133

PR #1039, opened 2022-11-29
fix: DEV-3867: Fix RelationsOverlay with 3391 ff on

hlomzik:


nicholasrq (approved):

src/components/RelationsOverlay/RelationsOverlay.js
```diff
@@ -265,7 +267,11 @@ let readinessTimer = null;
 const checkTagsAreReady = (tags, callback) => {
   clearTimeout(readinessTimer);
 
-  if (!isAlive(tags)) return;
+  if (isFF(FF_DEV_3391)) {
+    if (![...tags.values()].every(isAlive)) return false;
```
bmartel, line 271: Is the value of tags consistently not a mst value when the FF is on?

bmartel (approved):

PR #1053, opened 2022-12-01
fix: DEV-3820: Universal readonly

nicholasrq:
This PR addresses all readonly issues in different combinations:

- readonly annotations
- readonly results in the annotation
- readonly mode in view-all (same as the readonly annotation)

In addition, this PR adds missing e2e tests for those scenarios and some e2e re-organization.

Addresses: DEV-3820, DEV-3819, DEV-3816


src/components/SidePanels/DetailsPanel/RegionItem.tsx
```diff
@@ -118,12 +118,12 @@ const RegionAction: FC<any> = observer(({
...
       <Elem name="group" mod={{ align: 'left' }}>
-        {!(region.readonly || region.locked || !region.editable) && entityButtons}
+        {!region.isReadOnly() && entityButtons}
       </Elem>
       <Elem name="group" mod={{ align: 'right' }}>
         <RegionActionButton
-          icon={region.editable ? <IconLockUnlocked/> : <IconLockLocked/>}
-          disabled={region.readonly}
+          icon={region.isReadOnly() ? <IconLockUnlocked/> : <IconLockLocked/>}
+          disabled={region.isReadOnly()}
```
bmartel, line 126: This appears to be only able to be toggled to locked but never back again?

public/index.html
```diff
@@ -22,6 +22,12 @@
       In current functionality we create history record if we pass a result field. It doesn't let make a comment on review. Because we decided to relate comments to history.
       -->
     <link rel="stylesheet" href="/styles/main.css">
+    <script>
+      // window.APP_SETTINGS = {
+      //   feature_flags: {},
+      //   feature_flags_default_value: true,
+      // }
+    </script>
```
bmartel, line 30: Is this needed at all, just saw this was commented out.

src/tags/control/Choices.js
```diff
@@ -248,14 +259,13 @@ const Model = types
     },
 
     fromStateJSON(obj) {
+      console.log('from state json');
```
bmartel, line 262:
```suggestion
```

src/components/TopBar/Controls.js
```diff
@@ -31,17 +31,17 @@ const controlsInjector = inject(({ store }) => {
...
   const historySelected = isDefined(store.annotationStore.selectedHistory);
-  const { userGenerate, sentUserGenerate, versions, results, editable: annotationEditable } = annotation;
+  const { userGenerate, sentUserGenerate, versions, results, editable } = annotation;
   const buttons = [];
 
   const [isInProgress, setIsInProgress] = useState(false);
 
   // const isReady = store.annotationStore.selected.objects.every(object => object.isReady === undefined || object.isReady);
-  const disabled = !annotationEditable || store.isSubmitting || historySelected || isInProgress; // || !isReady;
+  const disabled = !editable || store.isSubmitting || historySelected || isInProgress; // || !isReady;
```
bmartel, line 42: Should this not also be observing `annotation.isReadonly()` instead of `annotation.editable`?

bmartel (changes requested):

public/index.html
nicholasrq: We probably must enable it in the future for testing purposes. The LSE and LS work this way enabling all flags by default. This might help us catch errors earlier.

I disabled it for now as we don't have tests that cover all feature flags.

src/components/TopBar/Controls.js
nicholasrq: No, we never locked the submit/update buttons in readonly. I would leave it like this until we decide something else.

src/mixins/ReadOnlyMixin.js
```diff
@@ -0,0 +1,29 @@
...
+    },
+  }));
+
+export const ReadOnlyRegionMixin = types
+  .model('ReadOnlyRegionMixin', {
+    readonly: types.optional(types.boolean, false),
+  })
+  .views(self => ({
+    isReadOnly() {
+      return self.locked || self.readonly || self.annotation.isReadOnly() || (parent && (self.parent.isReadOnly() || self.parent.result?.isReadOnly()));
```
Gondragos, line 17:
```suggestion
      return self.locked || self.readonly || self.annotation.isReadOnly() || (self.parent && (self.parent.isReadOnly() || self.parent.result?.isReadOnly()));
```

src/tags/control/Choices.js
```diff
@@ -151,6 +151,10 @@ const Model = types
       return 'choice';
     },
 
+    isReadOnly() {
+      return self.result?.isReadOnly() || self.annotation.isReadOnly();
+    },
```
Gondragos, line 156: Shouldn't we use ReadOnlyControlMixin instead?

src/utils/feature-flags.ts
```diff
@@ -127,6 +127,7 @@ export const FF_DEV_1284 = 'fflag_fix_front_dev_1284_auto_detect_undo_281022_sho
 function getFeatureFlags() {
   return {
     ...(window.APP_SETTINGS?.feature_flags ?? {}),
+    [FF_DEV_3391]: false,
```
Gondragos, line 130: I'm not sure it should be here.

Gondragos (approved):

hlomzik (changes requested):

hlomzik (approved):

PR #1060, opened 2022-12-02
feat: DEV-3844: Add loading state webcomponent

bmartel:


src/lib/AudioUltra/Visual/Loader.ts
```diff
@@ -0,0 +1,48 @@
+export class Loader extends HTMLElement {
+  constructor() {
+    super();
+    this.attachShadow({ mode: 'open' });
+    if (!this.shadowRoot) return;
+
+    this.shadowRoot.innerHTML = `
+      <style>
```
yyassi-heartex, line 8: short term, this is fine - but shouldnt we move this to one of the css/styl/sass files? I'd prefer if we reduce the amount of css we have in our JS/TS files

yyassi-heartex (approved):
lgtm but would prefer a tweak for the css as mentioned below

src/lib/AudioUltra/Visual/Loader.ts
bmartel: this is shadowDom styling, I'd say this is the most portable format we can use. We have to remember this is the lib, LSF. So long term I was thinking we use shadowDom with css vars and css parts, moving the actual styling and values out to css in LSF or whatever consumer applications. And just have some sensible defaults internal of the lib.

PR #1069, opened 2022-12-05
fix: DEV-3826: Toggling Audio interfaces should correctly update the view

yyassi-heartex:


src/tags/object/AudioUltra/view.tsx
```diff
@@ -163,6 +163,16 @@ const AudioUltraView: FC<AudioUltraProps> = ({ item }) => {
...
         mediaType="audio"
+        toggleVisibility={useCallback((layerName: string, isVisible: boolean) => {
+          if (waveform.current) {
+            const layer = waveform.current?.getLayer(layerName);
+
+            if (layer) {
+              layer.setVisibility(isVisible);
+            }
+          }
+        }, [waveform])}
```
bmartel, line 174: since this is a ref, passing it as a dependency doesn't make sense. It is a stable reference and cannot be stale, nor trigger shallow comparison change. Probably just remove this from the deps array.

PR #1073, opened 2022-12-06
fix: DEV-3919: Fix overflow to show audioform settings and volume popups

juliosgarbi:


juliosgarbi:
This fix is not the best way to fix this, closing this PR to fix directly in label-studio

PR #1076, opened 2022-12-07
chore: DEV-3947: Add e2e tests coverage to reports

Gondragos:


nicholasrq (changes requested):
It currently screws the paths of the files by duplicating the `src/` part leading to wrong coverage report.

.github/workflows/cicd_pipeline.yml
```diff
@@ -25,27 +25,46 @@ jobs:
   build_bundle:
     name: "Build JS Bundle"
     if: github.event_name == 'push' || github.event.pull_request.draft == false
-    uses: heartexlabs/label-studio-frontend/.github/workflows/build_bundle.yml@master
+    uses: heartexlabs/label-studio-frontend/.github/workflows/build_bundle.yml@dev-3947/e2e-coverage
     with:
       sha: ${{ github.event.pull_request.head.sha || github.event.after }}
 
+  build_for_coverage:
+    name: "Build JS Bundle"
```
farioas, line 33: Use another name

codecov.yml
```diff
@@ -1,2 +1,2 @@
 fixes:
-  - "::src/"
+  - "/home/runner/work/label-studio-frontend/label-studio-frontend/::"
```
farioas, line 2: absolute path is not backward compatible with local run

.github/workflows/tests_coverage.yml
```diff
@@ -0,0 +1,44 @@
...
+          path: coverage/
+          
+      - name: "Download e2e coverage artifacts"
+        uses: actions/download-artifact@v3
+        with:
+          name: e2e-tests-coverage
+          path: coverage/
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3.1.1
```
farioas, line 40: if: ${{ !github.event.pull_request.head.repo.fork }}

.github/workflows/tests_coverage.yml
```diff
@@ -0,0 +1,44 @@
+name: "Send tests coverage"
+
+on:
+  workflow_call:
+    inputs:
+      sha:
+        required: true
+        type: string
+
+env:
```
farioas, line 10: Not used

farioas (dismissed):

.github/workflows/cicd_pipeline.yml
nicholasrq: it's different on the job level, here name is used as a group, so there'll be two jobs:
- Build JS Bundle / Build
- Build JS Bundle / Build Coverage

codecov.yml
nicholasrq: we don't care about local runs as we're not uploading anything to codecov from local machines. this is a CI only thing so we're not expecting local paths there

nicholasrq (approved):

PR #1090, opened 2022-12-09
fix: DEV-3934: Nested elements within a V3 Audio region are NOT updated properly upon selecting another region

yyassi-heartex:


src/tags/object/AudioUltra/view.tsx
```diff
@@ -81,16 +81,15 @@ const AudioUltraView: FC<AudioUltraProps> = ({ item }) => {
 
       const selectRegion = (region: Region|Segment, event: MouseEvent) => {
         const growSelection = event.metaKey || event.ctrlKey;
+        const itemRegions = item.annotation.regions.filter((r: any) => r.object.name === item.name);
```
hlomzik, line 84: `item.regs`

src/tags/object/AudioUltra/view.tsx
```diff
@@ -81,16 +81,15 @@ const AudioUltraView: FC<AudioUltraProps> = ({ item }) => {
...
 
-        if (!region.selected && !region.isRegion)
+        if (!growSelection || (!region.selected && !region.isRegion))
           item.annotation.regionStore.unselectAll();
 
         // to select or unselect region
-        item.annotation.regions.forEach((obj: any) => {
+        itemRegions.forEach((obj: any) => {
```
hlomzik, line 90: there is only one region, so we have to just find it:
```js
const itemRegion = itemRegions.find(obj => obj.id === region.id);
itemRegion && obj.annotation.regionStore.toggleSelection(itemRegion, region.selected);
```

hlomzik: and yes, even simplier: `const itemRegion = item.regs.find...`

hlomzik (changes requested):

hlomzik (approved):

PR #1096, opened 2022-12-12
feat: DEV-3873: Labeling UI improvements

yyassi-heartex:


yyassi-heartex:
we no longer need this

PR #1098, opened 2022-12-13
chore: DEV-3645: E2E for TimeSeries with huge data and overview manipulations

Gondragos:
will also fix problems with `NaN` and `Infinity` values in svg attrs


Gondragos:
Incorrect name of the commit

PR #1104, opened 2022-12-13
fix: DEV-3904: Allow to clear DateTime

hlomzik:
Also don't allow to submit required full DateTime with empty date.

### DIFF
--- a/src/tags/control/DateTime.js
+++ b/src/tags/control/DateTime.js
@@ -257,9 +257,16 @@ const Model = types
     },
 
     setDate(dateArray) {
-      self.day = dateArray[2];
-      self.month = dateArray[1];
-      self.year = dateArray[0];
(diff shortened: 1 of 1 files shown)


src/tags/control/DateTime.js
```diff
@@ -302,7 +303,7 @@ const HtxDateTime = inject('store')(
       const validDateArray = item.validDateFormat(value);
 
       setDateInputValue(value);
-      if (validDateArray) item.setDate(validDateArray);
+      if (!value || validDateArray) item.setDate(validDateArray);
```
Gondragos, line 306: I think here we need some clarification about `false[2] === undefined` 😅

Gondragos (changes requested):

Gondragos (approved):

PR #1107, opened 2022-12-13
fix: DEV-3941: Enable adjusting player head position using TimeBox

juliosgarbi:


src/components/SidePanels/DetailsPanel/RegionEditor.tsx
```diff
@@ -73,8 +73,7 @@ const RegionEditorComponent: FC<RegionEditorProps> = ({
           endTime={region.end}
           minTime={0}
           maxTime={region?._ws_region?.duration}
-          startTimeReadonly={true}
-          endTimeReadonly={true}
+          endTimeReadonly={false}
```
bmartel, line 76: is there any reason this is set to false? I see the others just removed these properties.

src/components/TimeDurationControl/TimeBox.tsx
```diff
@@ -87,23 +87,29 @@ export const TimeBox: FC<TimerProps> = ({
   const handleChangeInput = (e: React.FormEvent<HTMLInputElement>) => {
     let input = e.currentTarget.value;
 
-    input = input.replace(/[^0-9]+/g, '');
-
-    input = `${input.substr(0, 2)}:${input.substr(2, 2)}:${input.substr(4, input.length)}`;
-
-    input = input.substring(0, 9);
+    input = input.replace(/\D/g, '')
```
bmartel, line 90: Currently this does not allow for a user to backspace a value, and type in the new value of some portion. It is really hard to accurately input the times you need to. I think we need some improvements to the user experience here.

bmartel (changes requested):

src/components/SidePanels/DetailsPanel/RegionEditor.tsx
juliosgarbi: Brandon, endTimeReadonly was set to true by default, but it makes more sense for both to be false by default and we set true when we need. I already changed this!

juliosgarbi:
wrong pr

PR #1111, opened 2022-12-14
fix: DEV-3941: Enable adjusting player head position using TimeBox

juliosgarbi:


bmartel (approved):
Looks good, checked the following:

- times do not exceed upper or lower boundaries of total duration
- can easily edit directly any portion of the time string and it correctly updates the time position
- the format is correct when selecting into the input box
- blurring the input correctly retains the time in the correct format
- works for times in ranges HH:mm:ss:sss, mm:ss:sss and ss:sss correctly

PR #1113, opened 2022-12-14
chore: DEV-2715: Do not build chunks with period used as separator

bmartel:


yyassi-heartex (approved):

bmartel:
Not the correct solution, going to close this.

PR #1114, opened 2022-12-14
Implements Magic Wand tool for image segmentation

BradNeuberg:
This pull request implements a Magic Wand, making it possible to click in a region of an image a user is doing segmentation labeling on, drag the mouse to dynamically change flood filling tolerance, then release the mouse button to get a new labeled area. It is particularly effective at labeling broad, diffuse, complex edged objects, such as clouds, cloud shadows, snow, etc. in earth observation applications or organic shapes in biomedical applications.

A short video of the tool in action can be seen here: https://www.youtube.com/watch?v=XxjqBx_0-wk

Detailed notes on its implementation can be seen at the top of src/tools/MagicWand.js.

e2e tests still need to be written, these will be added to this PR once done.


BradNeuberg:
Note that `examples/image_magic_wand/annotations/1.json` points to a Google Cloud bucket for serving up some example images; before landing this we will want to copy these to your public bucket and change the path. Note that some CORS headers will also have to be set on that bucket in order for the Magic Wand to work from demo images served from it; those changes are documented in the START.md file.

examples/image_magic_wand/tasks.json
```diff
@@ -0,0 +1,16 @@
+[
+  {
+    "id": 1202,
+    "data": {
+      "image": [
+        "https://storage.googleapis.com/brad-neuberg-label-studio/opensource/label-studio/examples/images/magic_wand_scale_1_20200902_015806_26_2235_1B_AnalyticMS_00750_00750.jpg",
```
BradNeuberg, line 6: FYI these GCP buckets will have to be changed to `https://htx-misc.s3.amazonaws.com/opensource/label-studio/examples/images` and the assets uploaded there to reference from this example.

BradNeuberg:
My old code was again label-studio-frontend 1.6.0. I've rebased from `ls-release/1.7.0` and updated the magic wand to work against the latest code. I've opened a new PR to clean up history from this git surgery: https://github.com/heartexlabs/label-studio-frontend/pull/1117

PR #1130, opened 2022-12-29
fix: DEV-4075: Fix taxonomy items overlap on search

Gondragos:


yyassi-heartex (commented):
code is looking good - but im wondering why we have a feature flag for this since this is a fix - is there a situation where some would not want to have this fix in place?

e2e/plugins/featureFlags.js
```diff
@@ -0,0 +1,95 @@
...
+  let helper;
+
+  for (const helperName of supportedHelpers) {
+    if (Object.keys(helpers).indexOf(helperName) > -1) {
+      helper = helpers[helperName];
+    }
+  }
+
+  if (!helper) {
+    console.error('Feature flags is only supported in Playwright');
```
hlomzik, line 26: there is `supportedHelpers`, so it would be better to use it here with `join(', ')`

e2e/tests/taxonomy.test.js
```diff
@@ -0,0 +1,722 @@
...
+  async function checkOverlapAndGap(text1, text2) {
+    const bbox1 = await I.grabElementBoundingRect(AtTaxonomy.locate(AtTaxonomy.item).find('label').withText(text1));
+    const bbox2 = await I.grabElementBoundingRect(AtTaxonomy.locate(AtTaxonomy.item).find('label').withText(text2));
+
+    bbox1.y2 = bbox1.y + bbox1.height;
+    bbox2.y2 = bbox2.y + bbox2.height;
+
+    if (
+      bbox1.y < bbox2.y && bbox2.y < bbox1.y2
+      || bbox1.y < bbox2.y2 && bbox2.y2 < bbox1.y + bbox1.height
```
hlomzik, line 26: `bbox1.y + bbox1.height` is `bbox1.y2`
for consistency

hlomzik (approved):

PR #1138, opened 2023-01-05
feat: DEV-4034: Rendering performance improvements for large-duration audio

bmartel:
- [-] Tests for the changes have been added/updated (for bug fixes/features)
- [-] Docs have been added/updated (for bug fixes/features)

- [ ] Product design
- [x] Frontend

To support larger duration audio files > 5 hours.

**Decoding**
Audio files are now decoded using an FFMPEG based decoding library using Web Assembly. This decoding process is performed in chunks of 30 minute durations on the file, and the process itself happens from a WebWorker off the main thread. Every source file is run through an `AudioDecoder` (handles the decoding process between WebWorker and the FFMPEG lib) instance via the `AudioDecoderPool` (manages AudioDecoder instances via a Proxy), which allows reuse of the same decoder for sources in multiple tag cases and deduplication of work for same requests that happen within the same time period. This ensures the least amount of memory is retained between workloads, while optimizing for the common use cases in LabelStudio. It also continues to ensure memory is freed between decoding to ensure no memory leaks occur by marking it for removal, which will destroy and cleanup any instance(s) not in use after a 1 second grace period.

**Rendering**
Audio data is rendered now within a time frame (time slicing with generators and requestAnimationFrame), and yields back to the remainder of processes on the main thread to ensure that in worst case scenarios it does not freeze the UI. The reason for this particular change is to allow far zoomed out renders to be able to quickly cancel out stale, potentially longer running renders allowing only the most fresh to complete. This enables a more freely useable x-axis zoom, y-axis zoom and resize of the entire interface due to window/container resize at lower zoom levels.

In addition to the time slicing, an alternative approach is used to render the very large data points (large duration, zoom level is zoomed far out). It reuses the currently drawn buffer canvas and shifts it over by the movement direction, rendering the small slice of data that is now in view. This stabilizes the worst case scenarios and produces ~60fps on massive amounts of data (tested with ~200 million data points).

**Decoding**
Audio files are currently decoded using the web audio api `decodeAudioData`. This has many usage implications on any file that is > 20 minutes on most hardware, as the underlying implementation does not throw errors, and instead crashes the browser with usage that would cause the api itself to throw an exception. Memory usage is erratic and it is impossible to control when or if it garbage collects any previous decodes. It is also impossible to chunk decode, and all decoding happens on the main thread as it is not possible to use this api from a WebWorker.

**Rendering**
Audio data was rendered using no time slicing techniques, and would be overwhelmed by larger or more frequent requests. It uses a similar sampling technique which does produce good results, but is not enough to stabilize performance past ~10 million data points.

https://www.npmjs.com/package/audio-file-decoder `v2.3.0`

Yes. Vastly improves performance, but more importantly stabilizes and removes uncontrolled crashes from happening.

No

- FFMPEG wasm directly https://github.com/ffmpegwasm/ffmpeg.wasm.
- Decode the waveform data on the server using tools like https://github.com/bbc/audiowaveform, and load and render that on the client.

```
ff_front_dev_2715_audio_3_280722_short
```

_(check only one)_
- [x] No

_(check all that apply)_
- [-] e2e (reusing current tests)
- [ ] integration
- [ ] unit

Audio v3 (codename AudioUltra)


src/lib/AudioUltra/Media/WaveformAudio.ts
```diff
@@ -1,38 +1,59 @@
...
+    return this.decoder?.channelCount || 1;
+  }
+
+  get duration() {
+    return this.el?.duration ?? 0;
   }
 
   get sampleRate() {
-    return this.context?.sampleRate ?? 44100;
+    return this.decoder?.sampleRate || 44100;
```
yyassi-heartex, line 48: this is minor but can we create a constant that defines what the 44100 is?

yyassi-heartex (changes requested):
looking good overall - just one minor note for clarity and consistency to create a constant for a large value to help define what it is

src/lib/AudioUltra/Media/WaveformAudio.ts
bmartel: yep, good call

yyassi-heartex (approved):

src/lib/AudioUltra/Media/AudioDecoder.ts
```diff
@@ -0,0 +1,260 @@
...
+      AudioDecoder.cache.set(src, instance);
+
+      return instance;
+    }
+
+    super();
+
+    // only allow one cached decoder at a time to prevent memory leaks
+    // and limit the memory usage of the browser
+    if (AudioDecoder.cache.size > 0) {
```
hlomzik, line 51: Does it work well when there are two `Audio` tags in the config?

src/lib/AudioUltra/Media/AudioDecoder.ts
```diff
@@ -0,0 +1,260 @@
...
+    if (this.cancelled || this.sourceDecoded) return Promise.resolve();
+
+    info('decode:start', this.src);
+
+    // clear any previous decoding state
+    this.decodeId = Date.now();
+    this.decodingPromise =  new Promise(resolve => (this.decodingResolve = resolve as any));
+    this.decoderPromise = new Promise(resolve => (this.decoderResolve = resolve as any));
+
+    await getAudioDecoderWorker(DecodeAudioWasm, arraybuffer)
```
hlomzik, line 154: I think it's better not to mix Promise and async ways in one place?
```
try {
  const worker = await getAudioDecoderWorker(DecodeAudioWasm, arraybuffer);
  <Promise body>
} catch() { <can we skip errors?> }
<finally>
```

src/lib/AudioUltra/Media/AudioDecoder.ts
```diff
@@ -0,0 +1,260 @@
...
+        const currentOffset = durationOffset;
+
+        durationOffset += nextChunkDuration;
+
+        this.worker
+          .decodeAudioData(currentOffset, nextChunkDuration, {
+            multiChannel: options?.multiChannel ?? false,
+            ...options,
+          })
+          .then(chunk => {
```
hlomzik, line 253: `return` is redundant here, so can be simplified to `.then(resolve)`

src/lib/AudioUltra/Media/AudioDecoder.ts
```diff
@@ -0,0 +1,260 @@
...
+
+    const totalDuration = this.worker.duration;
+
+    let durationOffset = 0;
+
+    while (true) {
+      yield new Promise(resolve => {
+        if (!this.worker || this.sourceDecodeCancelled) return resolve(null);
+
+        const nextChunkDuration = Math.min(DURATION_CHUNK_SIZE, Math.max(0, totalDuration - durationOffset));
```
hlomzik, line 243: `clamp()`

src/lib/AudioUltra/Media/MediaLoader.ts
```diff
@@ -45,29 +67,48 @@ export class MediaLoader extends Destructable {
       return Promise.resolve(null);
```
hlomzik, line 67: primitives can be returned directly from async functions. here and further

src/lib/AudioUltra/Media/MediaLoader.ts
```diff
@@ -45,29 +67,48 @@ export class MediaLoader extends Destructable {
...
 
-        return this.decodeAudioData(xhr.response).then((buffer) => {
-          if (buffer) {
-            return playAudio(buffer);
-          }
-          return null;
+        const decodingPromise = this.decodeAudioData(req);
+
+        if (this.audio) {
+          const decoderPromise = this.audio.decoderPromise || Promise.resolve();
```
hlomzik, line 100: move `await` here: `const decoderPromise = await this.audio.decoderPromise;`

src/lib/AudioUltra/Media/MediaLoader.ts
```diff
@@ -45,29 +67,48 @@ export class MediaLoader extends Destructable {
...
         }
 
-        return this.decodeAudioData(xhr.response).then((buffer) => {
-          if (buffer) {
-            return playAudio(buffer);
-          }
-          return null;
+        const decodingPromise = this.decodeAudioData(req);
+
+        if (this.audio) {
```
hlomzik, line 99: is this block intented to go in parallel? if yes it's better to exlpicitely say this here

hlomzik: but given that we already passed `if (!this.audio)` check this one is redundant

hlomzik (changes requested):

src/lib/AudioUltra/Media/AudioDecoder.ts
bmartel: It requires a 2 stage wait, which we need to know outside of this which has resolved for both when to show the loader vs when to begin drawing. It is also used to track the promises that have already succeeded, avoiding any further wasted work.

src/lib/AudioUltra/Media/AudioDecoder.ts
bmartel: Currently if they are different sources, it would cancel one the first decode result. I'm going to adjust this to instead cleanup instances on destroy with a buffer period, so it doesn't throw away in-use work (ex. switching annotations, submitting annotations, view all etc.), but allows for multiple decoder results to exist in parallel. This will not play nicely with very large audio, but will be fine for moderate sized durations.

yyassi-heartex (approved):

hlomzik (approved):

PR #1142, opened 2023-01-09
fix: DEV-4087: Fix multiple labels in Outliner

hlomzik:
Labels assigned from multiple `Labels` tags lead to duplicated region's text in Details panel. Cleaned up the code related to labels and text displaying.

Improved Details panel appearance.

Also cleaned up connected places:
- unused WithStates mixin (`getOneColor` was not used also)
- commented out typo in `displayMode` param usage


src/components/SidePanels/DetailsPanel/DetailsPanel.styl
```diff
@@ -22,11 +22,14 @@
...
 
 .region-actions
-  margin 8px
-  height 36px
+  padding 8px
+  height 52px
   display flex
   align-items stretch
   justify-content space-between
+  position sticky
```
hlomzik, line 30: comment from @Gondragos: small usual regions now have a bit weird floating appearance of actions panel on scroll. possibly can be fixed with some `top: 50vh` to trigger it only on huge content regions

hlomzik: but this is only actual for multiple selected regions when the overall height is higher than screen. and it's not critical anyway

PR #1148, opened 2023-01-11
fix: DEV-3793: LSDV-964: Relative coords for Image regions

hlomzik:
## Goal

Regions should store relative coords and sizes internally to fix different issues, related to sizes recalculations, canvas resize, value precisions, etc. It should also simplify the code by removing some extra code we currently use for recalculations.

Now the only places with size recalculations are:

- screen to internal on interactions (in handlers)
- internal to canvas on render (and they are memoized)

Nothing should be changed from user perspective except for values in Details — there will be internal percentage values instead of useless current sizes on current canvas without any visible references.

## Important changes

- There are new methods in `Image` — `internalToCanvasX/Y(n)` and `canvasToInternalX/Y(n)` to do these recalculations based on current canvas size (this size can be changed by window/panels resize)
- Regions have `canvas*` getters to get their canvas size. With FF off they just return internal value in screen coords, so it’s safe to use these getters directly on render.
- Now Image’s `event()` calls specific events with two sets of coordinates — internal and canvas ones for better flexibility.
- Temporary models with spooky `*AbsoluteCoordsDEV3793` postfix created to store old versions of the code for easy code clean up.
- Nothing happens on `updateImageSize()` — just image sizes change and regions are redrawn; internal coords are untouched, so no history triggers; the only calculation during every region rerender is getters `canvas*()` that will re-memoize new value, this is smaller than previous amout of calculations needed. All that except for Brushes — they are still using it.
- Simplified serialization — just dump the values. Values untouched the whole annotation life if region was not touched, imported and exported directly.
- Always correct values — they can’t be corrupted because of wrong canvas size, we don’t need silly checks like `stageWidth > 1`; precision is the same during sessions, because values are not changed.

## Collateral Goodness

- Finally got rid of outdated and not used state methods (`fromStateJSON`, `toStateJSON`, `copyState`, etc); the idea of states was in old version of LSF and was refactored two years ago, but some methods left stale; so it's time to go.
- Numbers in Details panel are more relevant now — previously they showed screen size of regions which is useless, any window/panel resize could change them, and image screen size is unknown for user. Now they represent percentage sizes like in the export.
- Some bugs are fixed with Relations and Selection because of small errors
- Other small interface fixes

## Things left

- Next goal is to strictly separate internal calculations and on-screen interactions. That should be done by splitting code and full type coverage. Type aliases could be possibly used for this, something like `type InternalNumber = number; type CanvasNumber = number;` and these types should not be interchanged.


src/mixins/Regions.js
```diff
@@ -179,53 +121,6 @@ const RegionsMixin = types
         console.error('Region class needs to implement serialize');
       },
 
-      toStateJSON() {
```
bmartel, line 182: Just for clarity, can you explain the removal of `toStateJSON`?

bmartel (approved):
I think after reviewing this many times, I believe it should be good. I tested as many scenarios as I could think of, all were working correctly.

src/regions/EllipseRegion.js
```diff
@@ -105,25 +93,6 @@ const Model = types
...
-          const { stageWidth, stageHeight } = self.parent;
-
-          if (stageWidth && stageHeight) {
-            self.setPosition(self.x, self.y, self.radiusX, self.radiusY, self.rotation);
-          }
-          break;
-        }
-      }
       self.checkSizes();
       self.updateAppearenceFromState();
```
Gondragos, line 97: It looks like it won't do anything now.
```suggestion
```

src/tools/Brush.js
```diff
@@ -153,7 +137,7 @@ const _Tool = types
         brush.addPoint(Math.floor(x), Math.floor(y));
       },
 
-      mouseupEv(ev, [x, y]) {
+      mouseupEv(ev, _, [x, y]) {
```
Gondragos, line 140: We need the same changes at Eraser

Gondragos:
We also have to update calculation of coordinates for `LabelOnBox` and for relations.

src/mixins/Regions.js
Gondragos: It's some kind of legacy code. We do not use it at all.

Gondragos:
The `Selection` tool also needs to correct the coordinates in that way. It relates to the "Transforming of multiple regions" test scenario.

Gondragos:
And there is also problems with `defaultDimensions` value. It should be easy to solve for rectangles, but for example polygons are a little bit different.

Gondragos:
I would suggest to use this code for `LabelOnKP` at https://github.com/heartexlabs/label-studio-frontend/pull/1148/files#diff-6041bbca1459baa8f6f6c573d812367624911512d5b747b789b45a0ef924ff50R288-R294 this solution:
```js
  if (!isLabeling && !isTexting) return null;
  const x = item.parent.internalToScreenX(item.x);
  const y = item.parent.internalToScreenY(item.y);
  const width = Math.max(item.parent.internalToScreenX(item.width), 2);

  return (
    <LabelOnBbox
      // keypoints' width scaled back to stay always small, so scale it here too
      x={x + (width + 2) / item.parent.zoomScale}
      y={y + (width + 2) / item.parent.zoomScale}
```
'Cause right now labels for `KeyPoints` are wrong positioned

We can improve precision of KeyPointRegion box by adding
```js
get bboxCoordsScreen() {
      const x = self.parent.internalToScreenX(self.x);
      const y = self.parent.internalToScreenY(self.y);
      const width = Math.max(self.parent.internalToScreenX(self.width), 2) / self.parent.zoomScale;

      return {
        left: x-width,
        top: y - width,
        right: x + width,
        bottom: y + width,
      };
    },
```
to it.

Potentially you can more or less fix all Relations by replacing https://github.com/heartexlabs/label-studio-frontend/blob/fb-dev-3793/relative-coords/src/components/RelationsOverlay/BoundingBox.js#L113-L147 
with
```js
    case 'rectangleregion':
    case 'ellipseregion':
    case 'polygonregion':
    case 'keypointregion':
    case 'brushregion': {
      const bbox = region.bboxCoordsScreen;

      return bbox ? stageRelatedBBox(
        region,
        {
          x: bbox.left,
          y: bbox.top,
          width: bbox.right - bbox.left,
          height: bbox.bottom - bbox.top,
        },
      ) : DEFAULT_BBOX;
    }
```

And I'm not sure how and when it was broken, but relations can desapear on image zoom. This may help:
https://github.com/heartexlabs/label-studio-frontend/blob/fb-dev-3793/relative-coords/src/components/RelationsOverlay/BoundingBox.js#L80
```js
{ x:region.parent.canvasSize.width, y:region.parent.canvasSize.height },
```

src/regions/PolygonRegion.js
```diff
@@ -48,20 +74,24 @@ const Model = types
...
+        bottom: Math.max(bboxCoords.bottom, point.y),
+      }), {
         left: self.points[0].x,
         top: self.points[0].y,
         right: self.points[0].x,
         bottom: self.points[0].y,
       });
+
+      // recalc on resize
+      fixMobxObserve(self.parent.stageWidth, self.parent.stageHeight);
```
Gondragos, line 92: I'm not sure if it affects something

Gondragos (approved):

Gondragos (approved):

PR #1149, opened 2023-01-12
feat: DEV-4187: Select regions without 300ms lag

hlomzik:


bmartel (approved):
I tested a variety of configs, this seems to work extremely well over the cases I tested.

PR #1153, opened 2023-01-19
fix: DEV-2019: Sync video timeline seeker with frame position

bmartel:


e2e/fragments/AtVideoView.js
```diff
@@ -0,0 +1,78 @@
...
+    return I.grabElementBoundingRect(this._positionSelector);
+  },
+
+  /**
+   * Drag the element to the given position
+   * @param {BoundingClientRect} bbox
+   * @param {number} x
+   * @param {number} [y=undefined]
+   */
+  async drag(bbox, x, y) {
```
yyassi-heartex, line 49: this feels useful and reusable enough that it should be moved to e2e/helpers/MouseActions.js

yyassi-heartex (changes requested):
just a minor change for e2e + I think we have a minor off by 1 issue where the first step after the slider gets repositioned 1 px too early 
image

e2e/fragments/AtVideoView.js
bmartel: moved this to just use the underlying one present in MouseActions.

yyassi-heartex (approved):

PR #1156, opened 2023-01-19
fix: DEV-4178: Disable postpone button when no comments left

juliosgarbi:


bmartel (approved):

src/components/TopBar/CurrentTask.js
```diff
@@ -20,8 +20,10 @@ export const CurrentTask = observer(({ store }) => {
     && !isDefined(store.annotationStore.selected.pk)
     && !store.canGoNextTask
     && !store.hasInterface('review')
+    && (isFF(FF_DEV_4174) && store.commentStore.comments.length >= 1)
```
hlomzik, line 23: with ff off `canPostpone` will always be `false`, should be ternary

hlomzik (changes requested):

PR #1157, opened 2023-01-19
chore: DEV-4230: Add tests for Outliner

Gondragos:


bmartel (approved):
Nice. I noticed you added a couple similar helpers for bbox as I had added in another PR. I will reuse those 👍.

PR #1166, opened 2023-01-25
fix: LSDV-4532: Audio v3 visual data is offset to sound playback

bmartel:
- [-] Tests for the changes have been added/updated (for bug fixes/features)
- [X] Docs have been added/updated (for bug fixes/features)

- [ ] Product design
- [X] Frontend

To support larger duration audio files > 5 hours.
The audio v3 performance branch was out of sync when playing back audio, and what was visually rendered in the wave. As a result the code had to be reverted to the previous state of audio v3, which removed all performance improvements.

#### What does this fix?
Fixes an offset issue with the underlying decoder lib in use. The Audio was always off by 1 second to the playback source. Also changed the point at which the audio media source begins loading, and when to pause audio on disconnecting the source to ensure playback is as immediate as possible.
I now removed the initial loading deduplication from the Audio view which was based on annotation store properties being set, in favor of the natural cache of the decoders. It is now redundant and will cause more inconsistencies like sometimes not loading the waveform on fresh reloads of the page in quickview.

Reintroduces the audio v3 performance improvements, without the audio delay.

**Decoding**
Audio files are now decoded using an FFMPEG based decoding library using Web Assembly. This decoding process is performed in chunks of 30 minute durations on the file, and the process itself happens from a WebWorker off the main thread. Every source file is run through an `AudioDecoder` (handles the decoding process between WebWorker and the FFMPEG lib) instance via the `AudioDecoderPool` (manages AudioDecoder instances via a Proxy), which allows reuse of the same decoder for sources in multiple tag cases and deduplication of work for same requests that happen within the same time period. This ensures the least amount of memory is retained between workloads, while optimizing for the common use cases in LabelStudio. It also continues to ensure memory is freed between decoding to ensure no memory leaks occur by marking it for removal, which will destroy and cleanup any instance(s) not in use after a 1 second grace period.

**Rendering**
Audio data is rendered now within a time frame (time slicing with generators and requestAnimationFrame), and yields back to the remainder of processes on the main thread to ensure that in worst case scenarios it does not freeze the UI. The reason for this particular change is to allow far zoomed out renders to be able to quickly cancel out stale, potentially longer running renders allowing only the most fresh to complete. This enables a more freely useable x-axis zoom, y-axis zoom and resize of the entire interface due to window/container resize at lower zoom levels.

In addition to the time slicing, an alternative approach is used to render the very large data points (large duration, zoom level is zoomed far out). It reuses the currently drawn buffer canvas and shifts it over by the movement direction, rendering the small slice of data that is now in view. This stabilizes the worst case scenarios and produces ~60fps on massive amounts of data (tested with ~200 million data points).

**Decoding**
Audio files are currently decoded using the web audio api `decodeAudioData`. This has many usage implications on any file that is > 20 minutes on most hardware, as the underlying implementation does not throw errors, and instead crashes the browser with usage that would cause the api itself to throw an exception. Memory usage is erratic and it is impossible to control when or if it garbage collects any previous decodes. It is also impossible to chunk decode, and all decoding happens on the main thread as it is not possible to use this api from a WebWorker.

**Rendering**
Audio data was rendered using no time slicing techniques, and would be overwhelmed by larger or more frequent requests. It uses a similar sampling technique which does produce good results, but is not enough to stabilize performance past ~10 million data points.

https://www.npmjs.com/package/audio-file-decoder `v2.3.0`

Yes. Vastly improves performance, but more importantly stabilizes and removes uncontrolled crashes from happening.

No

- FFMPEG wasm directly https://github.com/ffmpegwasm/ffmpeg.wasm.
- Decode the waveform data on the server using tools like https://github.com/bbc/audiowaveform, and load and render that on the client.

```
ff_front_dev_2715_audio_3_280722_short
```

_(check only one)_
- [ ] No
- [X] Not sure (briefly explain the situation below)

This change is behind a feature flag, but due to early usage by beta customers it needs to work for them. It is too difficult to introduce another feature flag.

_(check all that apply)_
- [-] e2e (reusing current tests)
- [ ] integration
- [ ] unit

Audio v3 (codename AudioUltra)


Travis1282 (approved):

src/lib/AudioUltra/Media/AudioDecoder.ts
```diff
@@ -229,7 +229,7 @@ export class AudioDecoder extends Events<AudioDecoderEvents> {
 
     const totalDuration = this.worker.duration;
 
-    let durationOffset = 0;
+    let durationOffset = -1;
```
hlomzik, line 232: please add comment for such unusual initial value

hlomzik:
about breaking changes — looks like it introduced such change already and was reverted :)
so I'd check at least "Not sure" with explanations

yyassi-heartex (approved):

PR #1174, opened 2023-02-02
feat: LSDV-4583-2: Image list pagination

nicholasrq:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

LSDV-4583
LSDV-4583-2

Along with `valueList` support this PR brings pagination support to travel between images in the `valueList`

Image with `valueList` always displays only the first image in the list

None

No

No

None

`feat_front_lsdv_4583_multi_image_segmentation_short`

_(check only one)_
- [x] No

_(check all that apply)_
- [x] e2e
- [ ] integration
- [ ] unit

Image segmentation, Labeling Interface


src/common/Pagination/Pagination.tsx
```diff
@@ -48,7 +54,7 @@ export const Pagination: FC<PaginationProps> = forwardRef<any, PaginationProps>(
   };
 
   return (
-    <Block name="pagination" mod={{ size }}>
+    <Block name="pagination" mod={{ size, outline, align, noPadding }}>
```
hlomzik, line 57: looks like too much parameters for tag used just twice. if the problem is in different designs maybe it's better to get them in sync?

src/tags/object/Image.js
```diff
@@ -289,11 +291,20 @@ const Model = types.model({
   },
 
   get parsedValue() {
-    return parseValue(self.value, self.store.task.dataObj);
+    return self._parsedValue = self._parsedValue ?? parseValue(self.value, self.store.task.dataObj);
```
hlomzik, line 294: why do you need this temporary private vars here and in `parsedValueList`?

hlomzik (changes requested):
and comments from https://github.com/heartexlabs/label-studio-frontend/pull/1173

src/tags/object/Image.js
nicholasrq: those are not temp, it's cache

src/common/Pagination/Pagination.tsx
nicholasrq: i prefer to be explicit. i could of course use single param like "bare=true" or somethig that combines these three. yet, it won't describe itself and I'll have to add documentation. also, these parameters might be useful in other pagination cases

src/tags/object/Image.js
hlomzik: but that's the getter, so it should not be called randomly. and if `value` is changed for some reason (LSF re-init from DM for example) it should be recalculated.

nicholasrq: yeah, you're right

yyassi-heartex (approved):

examples/image_list/START.md
```diff
@@ -0,0 +1,32 @@
+
```
hlomzik, line 1: weird file, remove it please

hlomzik (approved):

PR #1191, opened 2023-02-13
fix: DEV-4081: Handle edge case related to CORS headers and AWS S3

BradNeuberg:
- [*] Commit message(s) and PR title follows the format `[fix|feat|ci|chore|doc]: TICKET-ID: Short description of change made` ex. `fix: DEV-XXXX: Removed inconsistent code usage causing intermittent errors`
- [*] Tests for the changes have been added/updated (for bug fixes/features)
- [*] Docs have been added/updated (for bug fixes/features)
- [*] Best efforts were made to ensure docs/code are concise and coherent (checked for spelling/grammatical errors, commented out code, debug logs etc.)
- [*] Self-reviewed and ran all changes on a local instance (for bug fixes/features)

_(check all that apply)_
- [ ] Product design
- [*] Frontend

_(link to issue, supportive screenshots etc.)_

#### What does this fix?
_(if this is a bug fix)_

AWS S3's CORS headers have a bug where if an image is requested in an S3 bucket initially from an HTML `` tag the HTTP response incorrectly leaves out the HTTP `Origin` header. Then, later on, if a JavaScript CORS request is made to this same image, Chrome will return the cached request without the `Origin` header, which means the CORS request will fail. This issue is described here: https://bugs.chromium.org/p/chromium/issues/detail?id=409090

The user facing impact of this bug is that if there is a gallery of multiple images to be labelled beneath the stage, the thumbnails will be shown correctly, but once you click into them you get a CORS error shown on the stage.

The fix is as follows: If the Magic Wand feature flag is on AND if the Label Studio Image tag has the `crossOrigin` attribute set AND if the src image is an AWS S3 bucket, then we do a trick to break the cache for the image by adding a `?v=1` timestap on it.

_(if this is a breaking or feature change)_

_(if this is a breaking or feature change)_

Breaking the image cache can potentially affect image loading performance. We break the image cache as focused as possible (only for magic wand feature flag, only when using crossOrigin attribute, and only for AWS S3 URLs). In addition, we make sure that we are storing the timestamp as a volatile value on the Image object, so that we don't mindlessly keep breaking the cache with a new timestamp each time React might redraw the screen. We only need to break the cache on the initial image load.

Unfortunately there are no other known workarounds to this AWS S3 issue - Amazon should fix the bug on their side, but it sounds like they've dropped the ball. This does not effect other cloud bucket providers, such as GCP.

The Magic Wand feature flag: `FF_DEV_4081 = 'fflag_feat_front_dev_4081_magic_wand_tool'`

_(check only one)_
- [*] Yes, and covered entirely by feature flag(s)
- [ ] No

_(check all that apply)_
- [*] e2e
- [ ] integration
- [ ] unit

Image segmentation labelling and Image gallery if multiple images are present to be labelled

### DIFF
--- a/src/tags/object/Image.js
+++ b/src/tags/object/Image.js
@@ -34,22 +34,22 @@ import { FF_DEV_3377, FF_DEV_3666, FF_DEV_4081, isFF } from '../../utils/feature
  * @name Image
  * @meta_title Image Tags for Images
  * @meta_description Customize Label Studio with the Image tag to annotate images for computer vision machine learning and data science projects.
- * @param {string} name                       - Name of the element
- * @param {string} value                      - Data field containing a path or URL to the image
- * @param {boolean} [smoothing]               - Enable smoothing, by default it uses user settings
- * @param {string=} [width=100%]              - Image width
- * @param {string=} [maxWidth=750px]          - Maximum image width
- * @param {boolean=} [zoom=false]             - Enable zooming an image with the mouse wheel
- * @param {boolean=} [negativeZoom=false]     - Enable zooming out an image
- * @param {float=} [zoomBy=1.1]               - Scale factor
- * @param {boolean=} [grid=false]             - Whether to show a grid
- * @param {number=} [gridSize=30]             - Specify size of the grid
- * @param {string=} [gridColor="#EEEEF4"]     - Color of the grid in hex, opacity is 0.15
- * @param {boolean} [zoomControl=false]       - Show zoom controls in toolbar
- * @param {boolean} [brightnessControl=false] - Show brightness control in toolbar
- * @param {boolean} [contrastControl=false]   - Show contrast control in toolbar
- * @param {boolean} [rotateControl=false]     - Show rotate control in toolbar
- * @param {boolean} [crosshair=false]         - Show crosshair cursor
+ * @param {string} name                         - Name of the element
+ * @param {string} value                        - Data field containing a path or URL to the image
+ * @param {boolean} [smoothing]                 - Enable smoothing, by default it uses user settings
+ * @param {string=} [width=100%]                - Image width
+ * @param {string=} [maxWidth=750px]            - Maximum image width
+ * @param {boolean=} [zoom=false]               - Enable zooming an image with the mouse wheel
+ * @param {boolean=} [negativeZoom=false]       - Enable zooming out an image
+ * @param {float=} [zoomBy=1.1]                 - Scale factor
+ * @param {boolean=} [grid=false]               - Whether to show a grid
+ * @param {number=} [gridSize=30]               - Specify size of the grid
+ * @param {string=} [gridColor="#EEEEF4"]       - Color of the grid in hex, opacity is 0.15
+ * @param {boolean} [zoomControl=false]         - Show zoom controls in toolbar
+ * @param {boolean} [brightnessControl=false]   - Show brightness control in toolbar
+ * @param {boolean} [contrastControl=false]     - Show contrast control in toolbar
+ * @param {boolean} [rotateControl=false]       - Show rotate control in toolbar
+ * @param {boolean} [crosshair=false]           - Show crosshair cursor
  * @param {string} [horizontalAlignment="left"] - Where to align image horizontally. Can be one of "left", "center" or "right"
  * @param {string} [verticalAlignment="top"]    - Where to align image vertically. Can be one of "top", "center" or "bottom"
  * @param {string} [defaultZoom="fit"]          - Specify the initial zoom of the image within the viewport while preserving it’s ratio. Can be one of "auto", "original" or "fit"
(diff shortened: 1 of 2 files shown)


src/tags/object/Image.js
```diff
@@ -297,10 +299,23 @@ const Model = types.model({
...
-    const value = self.parsedValue;
+    let value = self.parsedValue;
 
-    if (Array.isArray(value)) return value[self.currentImage];
-    return value;
+    if (Array.isArray(value)) value = value[self.currentImage];
+
+    if (!self.imageCrossOrigin) {
+      return value;
+    } else if (isFF(FF_DEV_4081) && /^https?:.*?s3\.amazonaws\.com\//g.test(value)){
```
bmartel, line 308: Is this strictly an issue with AWS S3? And are there any other cases beyond that can trigger this problem?

BradNeuberg: It's a known issue with just AWS S3 not implementing the CORS spec correctly.

BradNeuberg:
I've updated to use a much simpler caching key, `?v=1`, and confirmed that everything works fine.

BradNeuberg:
@hogepodge Any updates on this ticket? I think the work and all feedback to comments is essentially complete, I'd love to land this if possible soon to close out tracking tickets on Planet's side related to this work.

hogepodge:
@BradNeuberg I'll check in on Monday in the next planning meeting.

src/tags/object/Image.js
```diff
@@ -297,10 +297,22 @@ const Model = types.model({
...
-    const value = self.parsedValue;
+    let value = self.parsedValue;
+
+    if (Array.isArray(value)) value = value[self.currentImage];
 
-    if (Array.isArray(value)) return value[self.currentImage];
-    return value;
+    if (!self.imageCrossOrigin) {
+      return value;
+    } else if (isFF(FF_DEV_4081) && /^https?:.*?s3\.amazonaws\.com\//g.test(value)){
```
bmartel, line 306: Hey @BradNeuberg 👋 . There is one other issue I came across with regards to this check and it occurs if the URL is a presigned URL. This query param addition made like this will break any presigned URL. If the URL is an S3 URL and it does not have an existing query param present `X-Amz-Signature`, then this cache bust would be fine.

BradNeuberg: Thanks @bmartel, good catch. I don't have an AWS S3 environment setup for testing (I use Google Cloud at Planet for my development). Do you have an example presigned S3 URL you can provide for testing (you can change the key on it to some gibberish so you don't leak anything)? It sounds like I should update my regular expression to handle the S3 presigned URL case?

BradNeuberg: Actually this is a tricky issue you've identified. So if I have an signed S3 URL like the following:

https://somebucket.s3.amazonaws.com/some-object-key?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=AKIAIOSFODNN7EXAMPLE%2F20230329%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20230329T123456Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=abcdefghijklm1234567890EXAMPLE

Can I simply not add a cache bust parameter at the end? Will that invalidate the entire S3 signed URL? Note that doing the cache bust works for Google Cloud signed bucket URLs, which is what we use at Planet. Not sure if there's something subtly different about S3 signed ones.

bmartel: Probably a safer way to do this cache bust would be to check if it is an amazon s3 url and whether it is a signed url, then perform the cache bust.

```suggestion
    } else if (isFF(FF_DEV_4081) && /^https?:.*?s3\.amazonaws\.com\//g.test(value) && !(new URL(value)).searchParams.has('X-Amz-Signature')) {
```

BradNeuberg: Thanks @bmartel, I've incorporated your code suggestion and pushed it back to this PR. Note that I am unable to test signed AWS URLs.

hogepodge:
@BradNeuberg, it's sounding like we're going to fix this in an upcoming patch that doesn't require cache busting. I'll reach out with more info once I have it.

bmartel:
@BradNeuberg There is a big overhaul coming to url handling with respect to CORS and external urls/presigning, which the changes as far as the frontend repository goes will be tracked here: #1299. This should be merged very soon (looking to be in by early next week), just some housekeeping items to take care of there. The underlying issue presented in the MagicWand scenarios were able to be reproduced and fixed there.

BradNeuberg:
@bmartel It sounds like I can close this PR without merging due to the other fixes you all already did around CORS handling?

bmartel:
Hey @BradNeuberg that is correct. This should now be fixed by the other PRs surrounding CORS and pre-signed url changes.

PR #1193, opened 2023-02-14
fix: LSDV-3009: Fix draft saving stuck

Gondragos:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

The problem with saving a draft with empty results affects the functionality of comments.

#### What does this fix?
This allows saving drafts with empty results. So that there should not be a problem with adding comments caused by this issue. And by this we also avoid draft saving permanent state. 

We can save a draft with empty results. But this solve the issue.

Currently saving a draft with empty results gets and this causes adding comments functionality to hang.

N/A

Nope

Nope

We also can rollback saving a draft if we can't save it. But this will create an ugly artifact in the UI and adding comments will still save an empty draft.

`fflag_fix_font_lsdv_3009_draft_saving_stuck_130223_short`

_(check only one)_
- [x] No

_(check all that apply)_
- [x] e2e
- [ ] integration
- [ ] unit

`Annotation History`, `Autosave`, `Draft`, `Comments`

### DIFF
--- a/src/stores/Annotation/Annotation.js
+++ b/src/stores/Annotation/Annotation.js
@@ -11,7 +11,16 @@ import Types from '../../core/Types';
 import Area from '../../regions/Area';
 import Result from '../../regions/Result';
 import Utils from '../../utils';
-import { FF_DEV_1284, FF_DEV_1598, FF_DEV_2100, FF_DEV_2100_A, FF_DEV_2432, FF_DEV_3391, isFF } from '../../utils/feature-flags';
+import {
+  FF_DEV_1284,
+  FF_DEV_1598,
+  FF_DEV_2100,
+  FF_DEV_2100_A,
+  FF_DEV_2432,
+  FF_DEV_3391,
+  FF_LSDV_3009,
+  isFF
+} from '../../utils/feature-flags';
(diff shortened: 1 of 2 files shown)


src/stores/Annotation/Annotation.js
```diff
@@ -647,7 +656,7 @@ export const Annotation = types
       const result = self.serializeAnnotation({ fast: true });
       // if this is new annotation and no regions added yet
 
-      if (!self.pk && !result.length) return;
+      if (!isFF(FF_LSDV_3009) && !self.pk && !result.length) return;
```
bmartel, line 659: @nicholasrq I recall you had added this to fix another issue, wouldn't this just be bringing back the original issue by disabling it with feature flag?

bmartel: This code I think was added further back, not sure what the intent was, but I think we just need to make sure of what this guarded against. It may very well not be required, which adding this FF to start the process of removing it would then make sense.

bmartel (approved):

hlomzik (approved):

borisheartex (approved):

PR #1197, opened 2023-02-14
docs: Get rid of Ranker docs

makseq:


hlomzik:
`Ranker` tag started a new life with new set of docs!

PR #1198, opened 2023-02-14
docs: Add list example 

makseq:


hlomzik:
`List` tag was reworked and works different now.

PR #1201, opened 2023-02-17
fix: LSDV-3012: Syncable with new simple approach

hlomzik:
New Syncable mixin to provide new and very simple and nice approach:
- don't send event back
- sync the whole state (time, speed, playing)
- group tags into syncing groups
- lock current sync window for any back-stream events
- sync is pluggable thing, tag should work without sync methods (Audio is still tied with Sync)

All tags are muted except Audio tags or the tag triggered even if no Audio is presented. That's because Audio is the only tag with controllable volume. May be changed in the future.

Things left:
- [x] sync speed
- [x] sync play when audio restarted after reaching the end
- [ ] untie Audio to trigger internal methods internally (too hard)
- [x] sync time shortly after play? can be helpful for overloaded browser

_(check all that apply)_
- [x] Product design
- [x] Frontend

Different issues with current sync. Limitation of only two tags synced. Too much coupled code.

#### What does this fix?
All events have their origin.
Every first event supresses all other tags from sending events during "sync window".
The only events can be firing during sync window are events from the origin tag.
Events are propagated only to other tags, not back to origin.

Tags use `sync=<synced group name>` attribute to sync any number of tags. 

Tags use `sync=<other tag name>` attribute to sync two tags with each other.

None. New `SyncableMixin` was created.

Should not. Sync itself is easier, some call chains were simplified, so this should improve performance a little.

No

Just fix the errors, but current sync was too overcomplicated. Some issues with very coupled code are still there :(

All changes except fixes for audio/video players are wrapped into FF
`fflag_feat_front_lsdv_3012_syncable_tags_070423_short`

_(check only one)_
- [x] No

This approach introduces new way to refer to synced tags with synced group name, so all tags in the group will have the same `sync` attribute.
But bacward-compatibility is preserved and old-style cross-referenced `sync` attrs are also acceptable.

_(check all that apply)_
- [x] e2e
- [ ] integration
- [x] unit


src/tags/object/AudioUltra/model.js
```diff
@@ -5,15 +5,16 @@ import { guidGenerator, restoreNewsnapshot } from '../../../core/Helpers.ts';
 import { AnnotationMixin } from '../../../mixins/AnnotationMixin';
 import IsReadyMixin from '../../../mixins/IsReadyMixin';
 import ProcessAttrsMixin from '../../../mixins/ProcessAttrs';
-import { SyncMixin } from '../../../mixins/SyncMixin';
+// import { SyncMixin } from '../../../mixins/SyncMixin';
```
bmartel, line 8:
```suggestion
```

src/tags/object/Video/Video.js
```diff
@@ -107,121 +103,72 @@ const Model = types
...
-        // Audio v1,v2
-        else {
-          Super.triggerSyncPause();
-        }
-      },
+    registerSyncHandlers() {
+      ['play', 'pause', 'seek'].forEach(event => {
+        self.syncHandlers.set(event, self.handleSync);
+      });
+      // self.syncHandlers.set('speed', self.handleSyncSpeed);
```
bmartel, line 139: We still need this right?

src/tags/object/Video/Video.js
```diff
@@ -107,121 +103,72 @@ const Model = types
...
+    },
+  }))
+  .actions(self => {
+    return {
+      afterCreate() {
+        const framerate = Number(parseValue(self.framerate, self.store.task?.dataObj));
 
-      handleSyncSpeed(speed) {
-        self.speed = speed;
+        if (!framerate || isNaN(framerate)) self.framerate = '24';
```
bmartel, line 165: Why are we keeping framerate as a String?

bmartel (changes requested):

src/mixins/Syncable.ts
```diff
@@ -0,0 +1,140 @@
...
+  .volatile<SyncableProps>(() => ({
+    syncHandlers: new Map(),
+    syncManager: null,
+  }))
+  /* eslint-enable @typescript-eslint/indent */
+  .actions(self => ({
+    afterCreate() {
+      if (!self.sync) return;
+
+      // @todo support sync by other tags' names for backward compatibility
```
bmartel, line 111: I think this has still yet to be done for the existing use cases to work, but even the most common use case is Video Timeline Segmentation from our own templates, that Video <-> Audio sample config doesn't work currently.

src/tags/object/Video/Video.js
hlomzik: Sadly that's our current way to deal with tag attrs =(
They are always strings (except for booleans), so we have to convert them when we use them. This can be actually done in other ways, but that requires to fix some places in whole app, so not for now...

Gondragos:
To fix "Audio Read only Regions" test, as I understand we need to teach `AtAudioView` to work with `Audio V3` component. I can suggest to use something like
```ts
async moveRegion(regionId, offset = 30) {
    const regionPosition = await I.executeScript(({ regionId, stageBbox }) => {
      const region = Htx.annotationStore.selected.regions.find(r => r.cleanId === regionId);

      if (window.APP_SETTINGS?.feature_flags?.ff_front_dev_2715_audio_3_280722_short === true) {
        const wsRegion = region._ws_region;

        if (!wsRegion.inViewport) {
          return null;
        }
        const { height } = wsRegion.visualizer;

        return {
          x: (wsRegion.xStart + wsRegion.xEnd) / 2 + stageBbox.x,
          y: height / 2 + stageBbox.y,
        };
      } else {
        const element = region.getRegionElement();
        const rect = element.getBoundingClientRect();

        return {
          x: rect.x + rect.width / 2,
          y: rect.y + rect.height / 2,
        };
      }
    }, { regionId, stageBbox: this._stageBbox });

    if (!regionPosition) return;

    return I.dragAndDropMouse(regionPosition, {
      x: regionPosition.x + offset,
      y: regionPosition.y,
    });
  },
```
instead of this https://github.com/heartexlabs/label-studio-frontend/blob/56b98f917a2101272a7e0fb5012f3e1545b51130/e2e/fragments/AtAudioView.js#L115-L131

It's not super accurate approach but it should work for now.

And after that we may need to correct this part
https://github.com/heartexlabs/label-studio-frontend/blob/56b98f917a2101272a7e0fb5012f3e1545b51130/e2e/fragments/AtAudioView.js#L101
with somthing like that
```ts
const _ws = Htx.annotationStore.selected.names.get(tagName)._ws;
let wrapper;

if (window.APP_SETTINGS?.feature_flags?.ff_front_dev_2715_audio_3_280722_short === true) {
  wrapper = _ws.visualizer.wrapper;
} else {
  wrapper = _ws.container;
}
```
But i'm still not sure about these magic numbers.
https://github.com/heartexlabs/label-studio-frontend/blob/56b98f917a2101272a7e0fb5012f3e1545b51130/e2e/tests/readonly-tests/readonly-results/audio-results-readonly.test.js#L67

e2e/fragments/AtVideoView.js
```diff
@@ -25,6 +32,22 @@ module.exports = {
...
+    return locate(this._seekStepForwardSelector).inside(this.locateRootSelector());
+  },
+
+  seekStepBackwardSelector() {
+    return locate(this._seekStepBackwardSelector).inside(this.locateRootSelector());
+  },
+
+  playButtonSelector() {
+    return locate(this._playButtonSelector).inside(this.locateRootSelector());
+  },
```
Gondragos, line 45: I would say that it's locators instead of selectors. As well that the other ones that use `locate`. Just for consistency. But yeah it's not critical.

e2e/tests/sync/audio-paragraphs.test.js
```diff
@@ -102,26 +102,24 @@ Scenario('Check audio clip is played when using the new sync option', async func
...
   await AtAudioView.waitForAudio();
-
-  I.waitForDetached('loading-progress-bar', 10);
-
   await AtAudioView.lookForStage();
 
   AtSidebar.seeRegions(2);
 
-  const [startingAudioPlusTime, startingParagraphAudioTime] = await AtAudioView.getCurrentAudioTime();
+  const [{ currentTime: startingAudioPlusTime }, { currentTime: startingParagraphAudioTime }] = await AtAudioView.getCurrentAudio();
```
Gondragos, line 109: This method looks like very unpredictable. And AtAudioView fragment shouldn't know about paragraphs.

src/mixins/Syncable.ts
```diff
@@ -0,0 +1,170 @@
...
+   * No back-sync to origin of the event.
+   * During SYNC_WINDOW only events from origin are processed, others are skipped
+   * @param {SyncData} data state to sync between connected tags
+   * @param {string} event name of event, supplementary info, actions should rely on data
+   * @param {string} origin name of the tag triggered event
+   * @returns {boolean} false if event was suppressed, because it's inside other event sync window
+   */
+  sync(data: SyncData, event: SyncEvent, origin: string) {
+    // @todo remove
+    if (!this.locked || this.locked === origin) console.log('SYNC', { event, locked: this.locked, data, origin });
```
Gondragos, line 58: Perhaps it's not a bad idea to highlight this with this comment. I'm not sure when we plan to be remove it.

src/tags/object/Video/Video.js
```diff
@@ -107,129 +103,80 @@ const Model = types
...
-        }
-      },
+      if (data.speed) {
+        self.speed = data.speed;
+      }
 
-      handleSyncSpeed(speed) {
-        self.speed = speed;
-      },
+      video.currentTime = data.time;
```
Gondragos, line 157: It looks like it doesn't affect time at `Timeline`. At least during my investigation sync event from the audio doesn't affect displayed value.

hlomzik: it was mysteriously broken because of removed resubscribe to events. most luckily because of weird internal logic dependent on sync operations and local events triggering.

e2e/tests/sync/audio-paragraphs.test.js
hlomzik: We can have audio in Paragraphs with `audioUrl` provided so it really has full media tag inside, but it's not very clear, I agree. Even the fact that `getCurrentAudio` returns array of media was surprising. But imho this can be solved by one line of documentation right now, because I don't know where it would evolve in future.

bmartel (commented):
Couple more cleanup items to resolve, but this does work well now. 

Playback rate now visually updates (I pushed a commit to address this), and the remaining usability works with most of the cases I tried. 

The tests now almost entirely pass, there is one aspect that should be looked into with the tests for playbackRate fail testing the audio paragraphs' playbackRate syncing not working correctly.

**Items remaining to address:**

- Compatibility for sync and Audio 1 (Not sure what we are doing about that, but everything now is using the new syncable approach and Audio 1 is still using the old one.)
- Cleanup of old Sync files (depending on what we do with the compatibility with Audio 1)
- Cleanup of some comments/commented out code
- Check paragraphs playbackRate sync
- Get the last 2 assertions passing for the playbackRate sync tests in the `audio-video-paragraphs.test.js`.

This could probably be QA'd, but to avoid getting things reopened multiple times I would want to have at least the playbackRate tests passing.

bmartel (approved):

src/mixins/Syncable.ts
```diff
@@ -147,15 +151,18 @@ const SyncableMixin = types
       const notSuppressed = self.syncManager!.sync(data, event, self.name);
 
       if (notSuppressed && event === 'play') {
-        self.syncMuted(false);
+        // if that's audio, it should not be muted.
```
Gondragos, line 154: It may be it's better to bring the cause from the commit description here for providing the context of the problem.

e2e/fragments/AtAudioView.js
```diff
@@ -291,4 +339,19 @@ module.exports = {
...
+   * @param {string} [message] for assertion
+   * @param {{ strict: boolean }} [options] only option is `strict` for strict comparison
+   */
+  seeTimesInSync(time1, time2, message = '', options = {}) {
+    if (options.strict) {
+      assert.equal(time1, time2, message);
+    } else {
+      assert.equal(Math.abs(time1 - time2) < TIME_DIFF_THRESHOLD, true, message);
+    }
+  },
```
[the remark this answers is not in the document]
hlomzik: As far as I can see this would not help... Assertions break test and show no stack trace :(

Gondragos (approved):

Issue #536, opened 2022-03-11
Dynamic zoom y-axis


dflatow:
### Is your feature request related to a problem?

There are times where a labeler may want to “zoom” into a particular vertical range of a time series. Currently my understanding is that you can set the y range via:

[https://github.com/heartexlabs/label-studio-frontend/pull/168#issue-805102232](https://github.com/heartexlabs/label-studio-frontend/pull/168#issue-805102232)

```
On: don't scale y axis at all, always fit for all the values on channel

Off: scale to fit currently displayed data

Can be changed globally for TimeSeries and locally on Channel
```

### Describe the solution you'd like

Ability to do one or more of the following:

- set the y-min and y-max via direct numerical input.
- slider connected with a numerical output ( so that the user can see which value is actually selected with the slider)
- selectable magnification levels (e.g. 1x, 2x, 3x, etc.) via something like a radial button.

jrdalenberg:
This is exactly what I need too for labeling EMG in multiple channels at the same time. When everything is dynamic, it is super difficult to judge relative amplitude differences @makseq. 

I also started a thread on slack about this.

PR #1206, opened 2023-02-21
Feature/custom y axis

jrdalenberg:
This PR adds an `ylim` parameter to the TimeSeries tag and allows a custom change to the y-axis. 

I see it as a a work in progress to address #536. 

If anyone could help me figure out how to let a user define the y-axis, which is what I am working towards, please let me know. I could use the help since I mainly work with Python, which makes this a challenge.


jrdalenberg:
Another evening of a fearsome battle with react ;)

The range does not update the entire chart yet. It happens when the x-axis slider is changed. Anyone got a tip to let it update? 

Screenshot from 2023-02-22 23-47-46

Credits for the dual slider design should go to @JauJau088.
(https://github.com/JauJau088/react-dual-slider)

PR #1230, opened 2023-03-07
fix: LSDV-4676: Region corruption MIG

nicholasrq:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

MIG feature caused a major data corruption which prevented users from using the feature.

#### What does this fix?
By leveraging raw region results, LSF can now set image dimensions without loading actual images. Also, during the serialization phase, LSF will use existing regions' data when it's impossible to calculate data from an image.

When submitting an annotation, LSF will export existing region results instead of calculating from scratch. This is only applicable when a related image is not yet loaded.

We're always trying to export data based on the image, which is sometimes impossible.

None

No

No

None

Same as the MIG feature.

_(check only one)_
- [x] No

_(check all that apply)_
- [x] e2e
- [ ] integration
- [ ] unit

Labeling Interface, Image Segmentation

### DIFF
--- a/e2e/tests/image-list.test.js
+++ b/e2e/tests/image-list.test.js
@@ -1,7 +1,3 @@
-/* global Feature, Scenario, locate */
-
-const { initLabelStudio } = require('./helpers');
-
 Feature('Image list via `valueList`');
 
 const config = `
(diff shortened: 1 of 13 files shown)


src/stores/Annotation/Annotation.js
```diff
@@ -978,6 +977,23 @@ export const Annotation = types
...
+          const tag = tagNames.get(obj.to_name);
+
+          if (tag.type !== 'image') return;
+
+          const imageEntity = tag.findImageEntity(obj.item_index ?? 0); 
+
+          if (!imageEntity) return;
+
+          imageEntity.setNaturalWidth(obj.original_width);
+          imageEntity.setNaturalHeight(obj.original_height);
```
Gondragos, line 995: Why it shouldn't be the same for `"rectanglelabels"` ?

src/stores/Annotation/Annotation.js
```diff
@@ -1091,7 +1107,7 @@ export const Annotation = types
       if (typeof objAnnotation !== 'object') {
         objAnnotation = JSON.parse(objAnnotation);
       }
-
+
```
bmartel, line 1110:
```suggestion
```

Just a rogue space here.

src/stores/Annotation/Annotation.js
```diff
@@ -978,6 +977,23 @@ export const Annotation = types
         if (tagNames.has(obj.from_name) && tagNames.has(obj.to_name)) {
           res.push(obj);
         }
+        
+        // Insert image dimensions from result 
+        (() => {
+          if (!obj.type.endsWith('labels')) return;
```
hlomzik, line 983: They can be just a usual tool result, like 'rectangle'. So I'd look specifically for `original_width`, because if they are not here, which is possible it would also be bad.

src/stores/Annotation/Annotation.js
hlomzik, line 1110: linting

e2e/tests/image-list.test.js
```diff
@@ -16,13 +12,79 @@ const config = `
...
+      'height': 18.64406779661017,
+      'rotation': 0,
+      'rectanglelabels': [
+        'Moonwalker',
+      ],
+    },
+    'origin': 'manual',
+    'to_name': 'img',
+    'from_name': 'tag',
+    'item_index': 1,
```
hlomzik, line 60: I think for correct test there is should be more than 4 images (because displayed one + 3 next are preloaded) and at least one index should be 4+

hlomzik (changes requested):

src/tags/object/Image/Image.js
```diff
@@ -320,8 +318,18 @@ const Model = types.model({
...
-    return {
-      ...value,
+    // We're using raw region result instead of calulated one when
+    // the image data is not available (image is not yet loaded)
+    // As the serialization also happens during region creation,
+    // we have to forsee this scenario and avoid using raw result
+    // as it can only be present for already created (submitter) regions
+    const useRawResult = !currentImageEntity.imageLoaded && isDefined(region._rawResult);
+
+    console.log({ useRawResult });
```
hlomzik, line 328: console.log

src/stores/Annotation/Annotation.js
```diff
@@ -1137,6 +1153,14 @@ export const Annotation = types
           };
 
           area = createArea(areaSnapshot);
+
+          // store copy of the original result inside the area
+          // useful when you need to serialize a result without
+          // updating it from current/actual data
+          // For safety reasons this object is always readonly
+          Object.defineProperty(area, '_rawResult', {
```
hlomzik, line 1161: I'd wrap this in FF just in case. This may lead to data corruption if something will go wrong

hlomzik (changes requested):

e2e/tests/image-list.test.js
nicholasrq: yeah, I will add it separately. for now it covers the case. on the other hand, all cases with preloading enabled must be tested

hlomzik (approved):

PR #1239, opened 2023-03-14
fix: LSDV-1476: Fix selected=true for Taxonomy

hlomzik:
Now `<Choice selected="true">` will work inside Taxonomy as well.

_(check all that apply)_
- [ ] Product design
- [x] Frontend

`Choice` tag has `selected` bool attribute to define preselected options in config. But this didn't work for `Choice`s inside `Taxonomy`.

#### What does this fix?
Adds `preselectedValues` to `Taxonomy`, fix init of `Choice`s inside Shared Store in order for them to have proper values.

"Usual" value is preselected in new annotations in configs like this:
```
<Taxonomy name="items" toName="text">
  <Choice value="Usual" selected="true" />
  ...
</Taxonomy>
```

Choices in Taxonomy can't be preselected.

none

`updateValue()` is called for every Choice in Taxonomy at initialization, so it may affect performance, but slightly.

nope

`preselectedValues` are current standard for such things, so no alternatives here.
For `Choice` value simpler approach can be used to get `value`/`alias` directly, not via getter, but it requires duplication of value retrieval logic (array/string).

`ff_dev_2100_preselected_choices_250422_short` to enable preselected choices
`fflag_fix_front_dev_3617_taxonomy_memory_leaks_fix` made it harder

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [x] integration
- [ ] unit

Taxonomy, Dynamic Choices


src/utils/feature-flags.ts
```diff
@@ -278,6 +276,18 @@ export const FF_LSDV_4998 = 'fflag_fix_front_lsdv_4998_missed_dynamic_children_0
  */
 export const FF_PROD_309 = 'fflag_feat_front_prod_309_choice_hint_080523_short';
 
+// Taxonomy and Choices related FFs, also for Dynamic Children and User Labels
+export const FFS_CHOICES = [
```
juliosgarbi, line 280: Why are we exporting this array?

Gondragos (approved):

PR #1252, opened 2023-03-16
fix: LSDV-4776: Fix interpolating a video region rotation prop

Gondragos:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

[_(link to issue, supportive screenshots etc.)_](https://github.com/heartexlabs/label-studio/issues/3720)
Upside down regions can get wrong rotation prop interpolation while they go through 180deg bound.

#### What does this fix?
This should fix logic of interpolation of rotation value of video regions.

N/A

No

No

N/A

N/A

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [x] unit

Video, interpolation

### DIFF
--- a/src/regions/VideoRectangleRegion.js
+++ b/src/regions/VideoRectangleRegion.js
@@ -4,7 +4,8 @@ import NormalizationMixin from '../mixins/Normalization';
 import RegionsMixin from '../mixins/Regions';
 import Registry from '../core/Registry';
 import { AreaMixin } from '../mixins/AreaMixin';
-import { interpolateProp, onlyProps, VideoRegion } from './VideoRegion';
+import { onlyProps, VideoRegion } from './VideoRegion';
+import { interpolateProp } from '../utils/props';
 
 const Model = types
   .model('VideoRectangleRegionModel', {
--- a/src/utils/__tests__/props.test.js
+++ b/src/utils/__tests__/props.test.js
@@ -1,0 +1,105 @@
+/* global describe, test, expect, it */
+
+import { interpolateProp, normalizeAngle } from '../props';
+
+describe('normalizeAngle', () => {
+  it('returns the same value for angle in the interval (-180; 180]', () => {
+    expect(normalizeAngle(-179.8)).toBe(-179.8);
+    expect(normalizeAngle(-42)).toBe(-42);
+    expect(normalizeAngle(0)).toBe(0);
+    expect(normalizeAngle(90)).toBe(90);
+    expect(normalizeAngle(180)).toBe(180);
+  });
+  it('projects angles from the intervals [-360;-180] and (180;360] onto the interval (-180; 180]', () => {
+    expect(normalizeAngle(-360)).toBe(0);
+    expect(normalizeAngle(-181)).toBe(179);
+    expect(normalizeAngle(220)).toBe(-140);
+    expect(normalizeAngle(359)).toBe(-1);
+  });
+});
+
+describe('interpolateProp', () => {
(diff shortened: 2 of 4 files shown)


src/regions/VideoRectangleRegion.js
```diff
@@ -4,7 +4,8 @@ import NormalizationMixin from '../mixins/Normalization';
 import RegionsMixin from '../mixins/Regions';
 import Registry from '../core/Registry';
 import { AreaMixin } from '../mixins/AreaMixin';
-import { interpolateProp, onlyProps, VideoRegion } from './VideoRegion';
+import { onlyProps, VideoRegion } from './VideoRegion';
+import { interpolateProp } from '../utils/props';
```
nicholasrq, line 8: not sure if common utils is the right place for it. interpolation only needed for video as we have "animation" there

src/utils/__tests__/props.test.js
```diff
@@ -0,0 +1,105 @@
...
+      size: 4,
+      frame: 100,
+    };
+
+    expect(interpolateProp(objA, objB, 25, 'x')).toBe(-5);
+    expect(interpolateProp(objA, objB, 50, 'x')).toBe(0);
+    expect(interpolateProp(objA, objB, 75, 'x')).toBe(5);
+    expect(interpolateProp(objA, objB, 8, 'y')).toBe(-86);
+    expect(interpolateProp(objA, objB, 30, 'y')).toBe(-185);
+    expect(interpolateProp(objA, objB, 72, 'y')).toBe(-374);
```
nicholasrq, line 60: here and below magic numbers all over the place. if we ever need to go back to this code, it will be impossible to remember why those are here and why we expect them. please, add some info here

also good to add docs to the `interpolateProps` itself

src/utils/props.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export const interpolateProp = (start: SequenceItem, end: SequenceItem, frame: number, prop: string) => {
+  // @todo edge cases
+  const r = (frame - start.frame) / (end.frame - start.frame);
+
+  if (prop === 'rotation') {
+    return normalizeAngle(start[prop] + normalizeAngle(end[prop] - start[prop]) * r);
+  }
+  return start[prop] + (end[prop] - start[prop]) * r;
+};
```
nicholasrq, line 25: please, add more info about the algorithm and provide some examples of what it takes and returns, and why

also, add explanation for the "edge cases"

src/utils/props.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * project any angle onto the interval (-180;180]
+ */
+export function normalizeAngle(x: number) {
+  while (x > 0) x -= 360;
+  return (x - 180) % 360 + 180;
+}
```
nicholasrq, line 7: it's considered a bad practice to reassign arguments' values. please, define a separate mutable variable for storing `x` instead

nicholasrq (changes requested):

src/regions/VideoRectangleRegion.js
Gondragos: We have the same common utils like `selection-tools` or `magic-wand`. And it still can be reused later for animations of something. It's edge case but I prefer to take it in account.

nicholasrq: makes sense, but we should be careful about putting stuff in common utils anyways. let's keep it for now though

nicholasrq (approved):

PR #1254, opened 2023-03-16
fix: LSDV-4774: Drawing on the hidden image in MIG scenario

nicholasrq:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

LSDV-4774

#### What does this fix?
Fixes an issue when it's possible to start a Brush/Polygon region on one image and continue drawing on a different image.

Regions created with Brush or Polygon will stay on the image they were initially created on. The user can continue drawing a region if they return to the image where they started. Otherwise, any drawings are ignored.

If a user creates an unfinished Polygon and switches to the next image, the Polygon will travel to the next image as well. Whenever the Polygon is finished, it will stay on the last selected image.
If a user creates a Bush region, switches to the next image, and starts drawing, they won't see anything. As soon as they switch back to the previous image, they will see everything they drew on the second image. Effectively, they were drawing on an invisible image.

None

No

No

None

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

Labeling UI, Image Segmentation, MIG

### DIFF
--- a/src/tools/Brush.js
+++ b/src/tools/Brush.js
@@ -182,9 +182,15 @@ const _Tool = types
         )
           return;
         const c = self.control;
+        const o = self.obj;
+
+        brush = self.getSelectedShape;
+
+        // prevent drawing when current image is
+        // different from image where the brush was started
+        if (o && brush && o.multiImage && o.currentImage !== brush.item_index) return;
 
         // Reset the timer if a user started drawing again
-        brush = self.getSelectedShape;
         if (brush && brush.type === 'brushregion') {
           self.annotation.history.freeze();
           self.mode = 'drawing';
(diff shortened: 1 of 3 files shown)


src/tools/Brush.js
```diff
@@ -182,9 +182,15 @@ const _Tool = types
...
           return;
         const c = self.control;
+        const o = self.obj;
 
-        // Reset the timer if a user started drawing again
         brush = self.getSelectedShape;
+
+        // prevent drawing when current image is
+        // different from image where the brush was started
+        if (o && brush && o.multiImage && o.currentImage !== brush.item_index) return;
```
Gondragos, line 191: Here is the same problem with `brush.item_index===null`. I'm not sure why it's happening yet.

Gondragos (changes requested):

Gondragos:
For `Brush` it may be that problem is here https://github.com/heartexlabs/label-studio-frontend/blob/68aeda880c17508f538252c705367d54e92850f6/src/tools/Brush.js#L105-L117
I think that `item_index` is missed there

Gondragos:
And for `Polygon` it maybe should be there https://github.com/heartexlabs/label-studio-frontend/blob/68aeda880c17508f538252c705367d54e92850f6/src/tools/Polygon.js#L115-L124

PR #1257, opened 2023-03-17
fix: LSDV-4747: Selecting the end character on a Paragraph phrase to the very start of other phrases only includes the first phrase selection

bmartel:
- [X] Tests for the changes have been added/updated (for bug fixes/features)

_(check all that apply)_
- [ ] Product design
- [X] Frontend

When selecting multiple lines (2+ phrases) in Paragraphs, if the selection was to end on the very beginning (zeroth character) of a subsequent phrase, the highlighted portion of the selection would only include the characters from the starting line of the selection.

#### What does this fix?
Ensures that selections include all lines within the selection, encompassing only valid phrase text and no other text that may happen to be within the range.

_(check only one)_
- [X] No

_(check all that apply)_
- [X] e2e
- [ ] integration
- [ ] unit

Paragraphs


src/tags/object/Paragraphs/Paragraphs.js
```diff
@@ -286,14 +302,56 @@ class HtxParagraphsView extends Component {
...
+   * @param {number} endOffset
+   * @returns {string}
+   */
+  _getResultText(textClass, phrases, range, start, startOffset, endOffset) {
+    const nodesInRange = getTextNodesInRange(range);
+    const endNodeIndex = nodesInRange.length - 1;
+
+    const resultText = nodesInRange
+      .reduce((acc, _n, i) => {
+        const str = phrases[+start + i]
```
hlomzik, line 324: I'm not sure here... This implies, that these returned text nodes will be perfectly aligned with phrases, but that doesn't feel true, because if we are filtering them out there are more text nodes than phrases in the range.

hlomzik: We are clearly looking here for text content of the whole phrase elements, so we can just calculate indices of start and end phrase, collect their texts and cut off by start/endOffsets. With no NodesInRange at all, because we actually don't need them in such easy case as Paragraphs.

hlomzik: Am I missing something?

hlomzik (changes requested):

src/tags/object/Paragraphs/Paragraphs.js
bmartel: Nope you are correct, this is now updated to using the same phraseElements, and simplified to just reuse the indices we already have.

hlomzik (approved):

PR #1273, opened 2023-03-25
fix: LSDV-4836: Conditional annotation with visibleWhen choice-selected works with Taxonomy

bmartel:

_(check all that apply)_
- [ ] Product design
- [X] Frontend

Make visibleWhen work with Taxonomy.

#### What does this fix?

Currently Taxonomy does not support visibleWhen, which throws errors and does not work in the Labelling config.
Support visibleWhen in Taxonomy so that it can be used to control other Tag's visibility conditions.

_(check only one)_
- [X] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

Taxonomy


src/regions/Result.js
```diff
@@ -170,11 +170,11 @@ const Result = types
           });
 
           if (!result) return false;
-          if (choiceValues && !choiceValues.some(v => result.mainValue.includes(v))) return false;
+          if (choiceValues && !choiceValues.some(v => result.mainValue.flat().includes(v))) return false;
```
hlomzik, line 173: `flat()` is a bit dangerous here, because it may consider selected upper-level same name Taxonomy item enough to submit this related result while the exact needed item is not selected. Could be something like `mainValue.map(v => Array.isArray(v) ? v.at(-1) : v).includes(v)`

hlomzik (changes requested):

src/regions/Result.js
bmartel: Good call 👍

e2e/tests/taxonomy.test.js
```diff
@@ -722,3 +722,240 @@ Scenario('Taxonomy maxUsages', async ({ I, LabelStudio, AtTaxonomy }) => {
...
+    });
+    I.say('Should see the full paths');
+    AtTaxonomy.clickTaxonomy();
+    AtTaxonomy.seeSelectedValues(['One to three / Three', 'Four to seven / Four']);
+    I.seeElement('.ant-checkbox-checked [name=\'Eight\']');
+  });
+});
+
+Scenario('Taxonomy visibleWhen with aliases', async ({ I, LabelStudio, AtTaxonomy }) => {
+  const createConfig = ({ showFullPath = false, visibleWhen = 'choice-selected', whenChoiceValue = 'Four' } = {}) => `
```
hlomzik, line 846: I don't think this is valid config — aliases should replace value completely if given afaik, so referring to this exact Choice should be only by alias — `4`. Do we have other requirements?

hlomzik (changes requested):

e2e/tests/taxonomy.test.js
bmartel: It replaces the value completely, but the label is still based on the value prop. So selections and targeting will still be based on the label `Four` but values in results and elsewhere is `4`. The assertions are driven by the label text from what I can see. Is this not correct? I based this on the other tests with regards to aliases.

hlomzik (approved):

PR #1289, opened 2023-04-03
feat: LSDV-3025-9: Add the remaining filters to filter component

juliosgarbi:

_(check all that apply)_
- [x] Product design
- [x] Frontend

It is added the remaining filters to filter component, but it's not integrated to outliner yet.
It is being added: Is between, Is not between, regex and is empty filters

no

no

fflag_feat_front_dev_3873_labeling_ui_improvements_short

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [x] unit

### DIFF
--- a/src/components/Filter/FilterInterfaces.tsx
+++ b/src/components/Filter/FilterInterfaces.tsx
@@ -8,7 +8,7 @@ export interface FilterListInterface {
   field?: string | string[] | undefined;
   operation?: string | string[] | undefined;
   value?: any;
-  path: string;
+  path?: string;
 }
 
 export interface AvailableFiltersInterface {
--- a/src/components/Filter/FilterInterfaces.tsx
+++ b/src/components/Filter/FilterInterfaces.tsx
@@ -7,7 +7,7 @@ export interface FilterInterface {
 export interface FilterListInterface {
   field?: string | string[] | undefined;
   operation?: string | string[] | undefined;
-  value?: string | string[] | undefined;
+  value?: any;
   path: string;
 }
 
--- a/src/components/Filter/FilterRow.styl
+++ b/src/components/Filter/FilterRow.styl
@@ -9,7 +9,7 @@
     height 16px
 
   &__column
-    display flex;
+    display flex
     margin-right 8px
 
     input
--- a/src/components/Filter/FilterRow.styl
+++ b/src/components/Filter/FilterRow.styl
@@ -9,6 +9,7 @@
     height 16px
 
   &__column
+    display flex;
     margin-right 8px
 
     input
--- a/src/components/Filter/__tests__/filter-utils.test.tsx
+++ b/src/components/Filter/__tests__/filter-utils.test.tsx
@@ -122,11 +122,101 @@ describe('FilterItems', () => {
     expect(filteredItems).toEqual(items);
   });
 
+  test('should filter items that have a greater value than the specified value ', () => {
+    const filterItem = { operation: 'greater', path: 'item.value', value: '25' };
+    const filteredItems = FilterItems(items, filterItem);
+
+    expect(filteredItems).toEqual([{
+      item: {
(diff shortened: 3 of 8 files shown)


src/components/Filter/FilterInterfaces.tsx
```diff
@@ -7,7 +7,7 @@ export interface FilterInterface {
 export interface FilterListInterface {
   field?: string | string[] | undefined;
   operation?: string | string[] | undefined;
-  value?: string | string[] | undefined;
+  value?: any;
```
yyassi-heartex, line 10: I think any will cause the linter to complain (warning not error) but aside from that i think this is fine - but are we expecting a `string | string[] | undefined` to come back?

src/components/Filter/FilterRow.styl
```diff
@@ -9,6 +9,7 @@
     height 16px
 
   &__column
+    display flex;
```
yyassi-heartex, line 12:
```suggestion
    display flex
```

src/components/Filter/FilterInterfaces.tsx
juliosgarbi: Probably no, this value can be boolean, string, different types of object, number, basically it can be anything as this value would be the value that would filter the content.

yyassi-heartex: ahh i see but since the user is inputing their value using a textbox - wouldnt it always start out as a string? or are is the value being cast to something else by the time it gets into `value`?

src/components/Filter/__tests__/filter-utils.test.tsx
```diff
@@ -150,4 +218,5 @@ describe('FilterItems', () => {
       },
     ]);
   });
+
```
yyassi-heartex, line 221:
```suggestion
```

src/components/Filter/filter-util.ts
```diff
@@ -120,6 +128,51 @@ const notequal = (items: any[], filterItem: FilterListInterface) => {
   }
 };
 
+const between = (items: any[], filterItem: FilterListInterface) => {
+  if (filterItem.value) {
+    return items.filter((obj) => {
+      const item = getFilteredPath(filterItem.path, obj);
+
+      return item >= filterItem.value.min && item <= filterItem.value.max;
```
yyassi-heartex, line 136: for easier readying can we change to this?
```suggestion
      return filterItem.value.min <= item && item <= filterItem.value.max;
```

src/components/Filter/FilterInterfaces.tsx
juliosgarbi: Not every time, for exemple when user selects if the value is between or not between we are calling the `RangeInput` from Number.js and in this case the value returns as an object `{max: 10, min: 1}` the same with is empty, that the value is returning as a boolean `true | false`.

src/components/Filter/filter-util.ts
```diff
@@ -120,6 +128,51 @@ const notequal = (items: any[], filterItem: FilterListInterface) => {
...
+    return items;
+  }
+};
+
+const notbetween = (items: any[], filterItem: FilterListInterface) => {
+  if (filterItem.value) {
+    return items.filter((obj) => {
+      const item = getFilteredPath(filterItem.path, obj);
+
+      return item <= filterItem.value.min || item >= filterItem.value.max;
```
yyassi-heartex, line 148: like wise here
```suggestion
      return item <= filterItem.value.min || filterItem.value.max <= item;
```

yyassi-heartex (approved):

src/components/Filter/FilterInterfaces.tsx
yyassi-heartex: ahh ok

PR #1297, opened 2023-04-06
fix: LSDV-4620-3: Fix Richtext performance and functional

Gondragos:
This is rethinking of the RichText engine.

This PR will:
1. Optimize the process of initialization of `RichText` which is helpful for work with large documents as a content and a lot of regions.
2. Allow working with hidden regions in `RichText` now hidden regions prevent interactions.
3. Fix edge cases of creating nested spans. For text like `abcd` right now you cannot define regions `ab`, `abcd` and then `cd`. Visually it still will be `ab`, `abcd` and `abcd` again. 
4. Fix critical memory leaks

_(check all that apply)_
- [ ] Product design
- [x] Frontend

The main reason is problems with performance in case of huge amount of content in data and a lot of created regions.
The sub reasons are bugs in creating highlighting spans and inability to work with parts of content that contain hidden regions.

N/A

It defenetly should. This thing must be 10x times faster in some cases.

I hope it does not.

There were options to rethink each part separately, but they are not very effective and not very accurate.

`fflag_feat_front_lsdv_4620_richtext_opimization_060423_short`
`fflag_fix_front_lsdv_4620_memory_leaks_100723_short`

_(check only one)_
- [ ] No
I mean it can if there are some cases that were missed. But we have some basic tests and the problem may be only with available functionality but as I hope not with data corruption.

_(check all that apply)_
- [ ] e2e
- [x] integration
- [ ] unit

`RichText`


e2e/tests/rich-text/rich-text-perfomance.test.js
```diff
@@ -0,0 +1,127 @@
...
+  };
+  const words = [];
+  const randomInt = createRandomIntWithSeed(seed);
+  let length = 0;
+  let paragraphWords = randomInt(0, 150);
+
+  let currentWordLength;
+
+  while ((currentWordLength = Math.min(randomInt(MIN_WORD_SIZE, MAX_WORD_SIZE), sybolsNum - length)) > 0) {
+    words.push(genWord(currentWordLength));
```
bmartel, line 36: Is this performance at all based on line length, not just the number of segments? I ask only because this test even though deterministically random, won't necessarily produce large line lengths.

src/tags/object/RichText/model.js
```diff
@@ -218,31 +268,85 @@ const Model = types
...
-
-        // node texts can be only retrieved from the visible node
-        self.regs.forEach(region => {
-          try {
-            region.updateHighlightedText();
-          } catch (err) {
-            console.error(err);
-          }
-        });
+        if (!isFF(FF_LSDV_4620_3)) {
```
bmartel, line 271: Maybe switch the if/else around so it reads in the affirmative case (have `if (isFF(FF_LSDV_4620_3)) {` come first), especially given this new behaviour we are hoping to make the official and only one, and it should be read first anywhere we are looking at this code.

src/tags/object/RichText/domManager.ts
bmartel, line 1: This appears like a Virtual DOM, specific to spans. Interesting 🤔

src/tags/object/RichText/domManager.ts
```diff
@@ -0,0 +1,624 @@
...
+    return this.elements.findIndex(el => {
+      return (el instanceof DDDynamicBlock) && el.start <= pos && el.end >= pos && el[avoid] !== pos;
+    });
+  }
+  getText(start: number, end: number) {
+    const startIdx = this.indexOfTextBlock(start, 'end');
+    const endIdx = this.indexOfTextBlock(end, 'start');
+
+    return this.elements.slice(startIdx, endIdx + 1).map(el => {
+      return typeof el !== 'string' ? el.getText(start, end) : el;
```
bmartel, line 377: We are sure that all other elements within the DOM representation, when calling their getText, that it returns a primitive value guaranteed?

bmartel (changes requested):

e2e/tests/rich-text/rich-text-perfomance.test.js
Gondragos: As I see this problem it's more about number of regions and the whole size of content. In the test I just tried to emulate natural variability of results, just in case.

src/tags/object/RichText/domManager.ts
Gondragos: I think we can guarantee it as all the elements here are instances of `DDDynamicBlock` wich in the end just makes concatenation of all values returned by all descendants that are `DDTextElement` instances which just slicing its string represetantion collected before.
And I think I now understand exactly what I should describe about `domManager`.

e2e/tests/rich-text/rich-text-perfomance.test.js
bmartel: That's fair, I think this does a good job of representing that.

src/tags/object/RichText/domManager.ts
```diff
@@ -0,0 +1,624 @@
...
+  }
+  removeNode() {
+    const { node } = this;
+    const doc = node.ownerDocument;
+    const parent = node.parentNode as Node;
+    const fragment = doc.createDocumentFragment();
+
+    while (node.firstChild) {
+      fragment.appendChild(node.firstChild);
+    }
```
bmartel, line 253: Can you add a comment here about what this is doing? I am not sure how this works at a glance and can only infer that appendChild mutates the value in some way that firstChild is iterated until null.

src/tags/object/RichText/domManager.ts
```diff
@@ -0,0 +1,624 @@
...
+  private container: HTMLDivElement|HTMLIFrameElement;
+  private root: HTMLBodyElement|HTMLDivElement;
+  private doc: Document;
+  private view: Window;
+  private domData: DomData;
+  private fragment: DocumentFragment;
+  private styleTags: {[key: string]: HTMLStyleElement};
+  private walker?:TreeWalker;
+  private currentPath:Path = new Path();
+  constructor(container:HTMLDivElement|HTMLIFrameElement) {
```
bmartel, line 450:
```suggestion

  constructor(container:HTMLDivElement|HTMLIFrameElement) {
```

src/tags/object/RichText/domManager.ts
```diff
@@ -0,0 +1,624 @@
...
+    return '';
+  }
+}
+
+class DomData {
+  private elements: Array<DDStaticElement|DDDynamicBlock|DDExtraText> = [];
+  private endPos: number;
+  private displayedText = '';
+  private displayedTextPos = 0;
+  constructor() {
```
bmartel, line 290:
```suggestion

  constructor() {
```

src/tags/object/RichText/domManager.ts
```diff
@@ -0,0 +1,624 @@
...
+    this.children.push(new DDTextElement(textNode, start, end, text, path));
+    this.end = end;
+  }
+}
+
+class DDStaticElement {
+  public node: HTMLElement;
+  public start: number;
+  public path: string;
+  constructor(node: HTMLElement, start: number, path: string) {
```
bmartel, line 275:
```suggestion

  constructor(node: HTMLElement, start: number, path: string) {
```

src/tags/object/RichText/domManager.ts
```diff
@@ -0,0 +1,624 @@
...
+      fragment.appendChild(node.firstChild);
+    }
+
+    parent.replaceChild(fragment, node);
+  }
+}
+
+class DDDynamicBlock extends DDBlock {
+  public path: string;
+  constructor(start: number, path: string) {
```
bmartel, line 261:
```suggestion

  constructor(start: number, path: string) {
```

src/tags/object/RichText/domManager.ts
```diff
@@ -0,0 +1,624 @@
...
+      }
+    }
+    checkStack();
+    this.children = result;
+  }
+}
+
+class DDSpanElement extends DDBlock {
+  public node: HTMLSpanElement;
+  constructor(node: HTMLSpanElement, start: number, end: number) {
```
bmartel, line 241:
```suggestion

  constructor(node: HTMLSpanElement, start: number, end: number) {
```

src/tags/object/RichText/domManager.ts
```diff
@@ -0,0 +1,624 @@
...
+    this.node.data += elements.map(el => el.node.data).join('');
+    this.end = elements[elements.length - 1].end;
+    this.text += elements.map(el => el.text).join('');
+  }
+}
+class DDBlock {
+  public start: number;
+  public end: number;
+  public children: Array<DDSpanElement|DDTextElement> = [];
+  constructor(start: number, end: number = start) {
```
bmartel, line 92:
```suggestion

  constructor(start: number, end: number = start) {
```

src/tags/object/RichText/domManager.ts
```diff
@@ -0,0 +1,624 @@
...
+
+    parent.removeChild(node);
+  }
+  mergeWith(elements:DDTextElement[]) {
+    this.node.data += elements.map(el => el.node.data).join('');
+    this.end = elements[elements.length - 1].end;
+    this.text += elements.map(el => el.text).join('');
+  }
+}
+class DDBlock {
```
bmartel, line 88:
```suggestion

class DDBlock {
```

src/tags/object/RichText/domManager.ts
```diff
@@ -0,0 +1,624 @@
+import { flatten, isDefined } from '../../../utils/utilities';
+
+type DDExtraText = string;
+class DDTextElement {
+  public node: Text;
+  public start: number;
+  public end: number;
+  public text: string;
+  public path?: string;
+  constructor(node: Text, start: number, end: number, text: string, path?: string) {
```
bmartel, line 10:
```suggestion

  constructor(node: Text, start: number, end: number, text: string, path?: string) {
```

bmartel (approved):
Just a bit of cleanup left but overall this should be looked at by QA now as it seems to be a good change. I'm not sure we have covered all of the possibilities in the tests here as I don't hold complete domain knowledge over RichText use cases, but there is a healthy representation covered and it seems reasonable from what I have seen.

src/tags/object/RichText/domManager.ts
Gondragos: That's exactly as you said. `appendChild` changes parent of `firstChild` wich moves it from it's previous parent's children list and changing of this list changes `firstChild` also untill all childrend leave the parent.

src/tags/object/RichText/domManager.ts
```diff
@@ -1,34 +1,51 @@
 import { flatten, isDefined } from '../../../utils/utilities';
 
+// line feed
+const LF = '\n';
```
bmartel, line 4:
```suggestion
const LINE_FEED = '\n';
```
It would be better to qualify the constant instead of adding a comment above the abbreviated as this is then clear at point of usage.

src/tags/object/RichText/domManager.ts
```diff
@@ -298,52 +333,100 @@ class DomData {
...
+      // that means that it contains some \n or other symbols that are trimmed by browser
+
+      // calc the offsets of the part of displayedText that matches the text in terms of displayed symbols
+      const { fromIdx, toIdx, content: newContent } = this.findProjectionOnDisplayedText(text);
+
+      pos = fromIdx;
+      displayedTextLength = toIdx - fromIdx;
+      // fill content with dummies for not displayable symbols
+      content = newContent;
+    }
```
bmartel, line 402: This is an important find, awesome work! I did notice this type of issue within Paragraphs as well, and we have a similar type of logic there. I think after seeing this work, we probably should look to reuse this DomManager there as well, and remove any unnecessary logic from the Paragraphs. Not to be done here, but a future improvement and code reuse.

bmartel (approved):

src/LabelStudio.js
```diff
@@ -62,20 +64,68 @@ export class LabelStudio {
...
+      if (isFF(FF_LSDV_4620_3_ML)) {
+        clearRenderedApp();
+      }
       destroySharedStore();
+      if (isFF(FF_LSDV_4620_3_ML)) {
+        /*
+           It seems that destroying children separately helps GC to collect garbage
+           ...
+         */
+        this.store.selfDestroy();
```
bmartel, line 116: This actually makes sense given what we know about MST and its underlying cache of nodes. The process I believe creates a dead set of cache nodes which won't release if attempting to destroy along with the references. So clearing then releasing everything remaining will allow for the Map values to be removed. It is unfortunate that we have to do this manually, but a really great find 🎉 .

src/LabelStudio.js
```diff
@@ -62,20 +64,68 @@ export class LabelStudio {
...
+      // cleanDomAfterReact needs this key to be sure that cleaning affects only current react subtree
+      const reactKey = findReactKey(childNodes[0]);
 
-    const destructor = () => {
       unmountComponentAtNode(rootElement);
+      /*
+        Unmounting doesn't help with clearing React's fibers
+        but removing the manually helps
+        @see https://github.com/facebook/react/pull/20290 (similar problem)
+        That's maybe not relevant in version 18
```
bmartel, line 91: Yea this will not be a problem once we move to React 18, and use the same compiled React throughout. In fact, the latter is likely what is snagging this now in React 17. Because we are operating with separate compiled React versions (even though they are the same version) it is not going to be referencing the identical code points, just an adjacent *similar* code point. That can only cause problems within the context of React's operations.

src/mixins/AreaMixin.js
```diff
@@ -40,7 +43,7 @@ export const AreaMixinBase = types
     },
 
     get texting() {
-      return self.results.find(r => r.type === 'textarea' && r.hasValue);
+      return isAlive(self) && self.results.find(r => r.type === 'textarea' && r.hasValue);
```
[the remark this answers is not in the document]
Gondragos: Semantically it should be `undefined`, yeah. But here I didn't care about returned value 'cause `isAlive` is here just to prevent errors on the stage of destroying storage nodes. I mean, we event will not see the result of rendering in that case.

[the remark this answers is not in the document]
Gondragos: But logically `hasValue` also returns `boolean`... And logically on the stage of destroying `false` is also OK 'cause there is still no texting.

src/mixins/HighlightMixin.js
```diff
@@ -120,11 +202,13 @@ export const HighlightMixin = types
...
       self.annotation.setHighlightedNode(self);
 
-      self.addClass(stateClass.active);
+      self.addClass(STATE_CLASS_MODS.active);
 
       const first = self._spans?.[0];
 
-      if (!first) return;
+      if (!first) {
+        return;
```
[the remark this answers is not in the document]
Gondragos: You got me. 😂 I actually use SonarLint now and it tells me what to do. I have to use `void 0` in all places where I use function/method with returning values, cause it should return something expectable and we should be able to read it from code. But this is a function/method without expected returning value. So I do not need to return somthing in it.

PR #1317, opened 2023-04-14
fix: LSDV-4881: Fix missed points in optimized timeseries

Gondragos:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

When the user zooms in with big time series (e.g. 250k points), some of the points are not selectable

#### What does this fix?
It makes available for the cursor all timestamps/points that was filtered by optimized data mode. 

N/A

N/A

N/A

It may be. But it was not detected by me.

Nope

N/A

`fflag_fix_front_lsdv_4881_timeseries_points_missing_140423_short`

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [x] integration
- [ ] unit

`TimeSeries`


src/tags/object/TimeSeries/Channel.js
```diff
@@ -543,6 +543,17 @@ class ChannelD3 extends React.Component {
...
+
+    if (isFF(FF_LSDV_4881)) {
+      originalSeries = series.filter(x => {
+        return x[column] !== null;
+      });
+      originalTimes = originalSeries.map(x => {
+        return x[time];
+      });
+    }
+
```
bmartel, line 556: Would this not need to completely replace the [series](https://github.com/heartexlabs/label-studio-frontend/pull/1317/files#diff-e50484b6cf864f3df51196bffc121e802d8a516aff1901cea68062d2eaee9322R562), [times](https://github.com/heartexlabs/label-studio-frontend/pull/1317/files#diff-e50484b6cf864f3df51196bffc121e802d8a516aff1901cea68062d2eaee9322R570), and [values](https://github.com/heartexlabs/label-studio-frontend/pull/1317/files#diff-e50484b6cf864f3df51196bffc121e802d8a516aff1901cea68062d2eaee9322R574) to be consistent?

bmartel: I am wondering about this because I think the `extent` would be incorrect. Plus if we can make this change holistically we would not have to differentiate [stickTimes](https://github.com/heartexlabs/label-studio-frontend/pull/1317/files#diff-e50484b6cf864f3df51196bffc121e802d8a516aff1901cea68062d2eaee9322R621)

Gondragos: From my point of view the functional of stick displaying is isolated enought to not touch anything else. The main purpose of this fix is to make every timestamp to be able to be displayed and selected. And it looks for me that it helps with that.
As I understand existing  `series`, `times`, `values` and `extent`  are part of the functional which may use optimization with skiping some values. If I return all timestamps there I will break this optimithation, and that's I defenetly do not want to do.

bmartel (approved):

PR #1327, opened 2023-04-19
fix: LSDV-4864: Magic Wand doesn't work with MIG 

nicholasrq:
### Description
When both MIG and image preloading are enabled, there was no canvas overlay for the Magic Wand which caused the Magic Wand to fail.

_(check all that apply)_
- [ ] Product design
- [x] Frontend

LSDV-4864

#### What does this fix?
Fixes inability to use the Magic Wand when image preloading is enabled

Magic Wand works with image preloading enabled

Magic Wand doesn't work when the image preloading is enabled

None

No

No

None

`fflag_feat_front_dev_4081_magic_wand_tool`
`fflag_feat_front_lsdv_4583_6_images_preloading_short`

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

Labeling Interface, Image Segmentation, Magic Wand

### DIFF
--- a/src/components/ImageView/Image.js
+++ b/src/components/ImageView/Image.js
@@ -11,6 +11,7 @@ export const Image = observer(forwardRef(({
   updateImageSize,
   usedValue,
   size,
+  overlay,
 }, ref) => {
   const imageSize = useMemo(() => {
     return {
@@ -26,6 +27,7 @@ export const Image = observer(forwardRef(({
 
   return (
     <Block name="image" style={imageSize}>
+      {overlay}
       <ImageProgress
         downloading={imageEntity.downloading}
         progress={imageEntity.progress}
@@ -57,7 +59,7 @@ const ImageProgress = observer(({
   return downloading ? (
     <Block name="image-progress">
       <Elem name="message">Downloading image</Elem>
-      <Elem tag="progress" name="bar" value={progress} min="0" max={1} step={0.0001}/>
+      <Elem tag="progress" name="bar" value={progress} min="0" max={1} step={0.0001} />
     </Block>
   ) : error ? (
     <ImageLoadingError src={src} value={usedValue} />
@@ -97,6 +99,6 @@ const ImageLoadingError = ({ src, value }) => {
   }, [src]);
 
   return (
-    <ErrorMessage error={error}/>
+    <ErrorMessage error={error} />
   );
 };
--- a/src/components/ImageView/ImageView.js
+++ b/src/components/ImageView/ImageView.js
@@ -468,7 +468,7 @@ const CanvasOverlay = observer(({ item }) => {
       }}
       style={item.imageTransform}
     />
-  ) : null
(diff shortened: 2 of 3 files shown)


bmartel (approved):

src/utils/canvas.js
```diff
@@ -198,6 +199,21 @@ function exportRLE(region) {
 
   document.body.appendChild(canvas);
 
+  if (region.maskDataURL) {
+    await new Promise((resolve) => {
+      const image = new Image();
+
+      image.onload = () => {
+        console.log('image loaded');
```
hlomzik, line 207: remove console.log

hlomzik (approved):

src/stores/AppStore.js
```diff
@@ -727,15 +726,20 @@ export default types
     }
 
     function setHistory(history = []) {
+      console.log('Setting annotation history');
       const as = self.annotationStore;
 
+      const selected = as.selected?.toJSON();
+
+      console.log({ pk: selected?.pk, id: selected?.id });
```
yyassi-heartex, line 734: are we looking to keep these console.logs temporarily? or are is this debug logs that were missed?
```suggestion
      const as = self.annotationStore;
      const selected = as.selected?.toJSON();
```

yyassi-heartex (approved):
there are a couple of very minor things that i wouldn't consider show stoppers but would be great to at least acknowledge if not accept suggestion

public/index.html
```diff
@@ -159,7 +159,7 @@
             }
           })
           ls.on("updateAnnotation", (_, annotation) => {
-            console.log(annotation.serializeAnnotation());
+            annotation.serializeAnnotation().then(r => console.log(r));
```
hlomzik, line 162:
```suggestion
            annotation.serializeAnnotation().then(console.log);
```

src/stores/AppStore.js
```diff
@@ -727,15 +726,20 @@ export default types
...
+
       as.clearHistory();
 
       // always check that history is for correct and submitted annotation
-      if (!history.length || !as.selected?.pk) return;
-      if (Number(as.selected.pk) !== Number(history[0].annotation_id)) return;
+      if (Array.isArray(history) && !history.length || !as.selected?.pk) return console.warn('Invalid history record', history);
```
hlomzik, line 739: why do we need check for array here? could it be something else and why?
we use it as array 2 lines down with no checks

hlomzik (approved):

src/stores/AppStore.js
nicholasrq: it's js, it could be anything :)

nicholasrq: I assume at some point we were getting "cannot read length of undefined" or something like that. Honestly, I don't remember already why I did that. Shouldn't be a problem though. At least this code shows that we expect an array and nothing else.

PR #1344, opened 2023-05-02
fix: LSDV-3025-12: Fully hide the dropdown component when closing it

juliosgarbi:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

The dropdown component is not hidden enterally, there is a small timeout that is making the transition method is being triggered before afterTransition. And it is making the dropdown is being clickable even after it's hidden.

no

no

na

na

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [x] unit

### DIFF
--- a/src/common/Dropdown/DropdownComponent.tsx
+++ b/src/common/Dropdown/DropdownComponent.tsx
@@ -23,6 +23,7 @@ export interface DropdownProps {
   enabled?: boolean;
   inline?: boolean;
   className?: string;
+  dataTestId?: string;
   style?: CSSProperties;
   children?: JSX.Element;
   onToggle?: (visible: boolean) => void;
@@ -79,8 +80,10 @@ export const Dropdown = forwardRef<DropdownRef, DropdownProps>(({
             setVisibility(visible ? 'before-appear' : 'before-disappear');
           },
           afterTransition: () => {
-            setVisibility(visible ? 'visible' : null);
-            resolve();
+            setTimeout(() => {
+              setVisibility(visible ? 'visible' : null);
+              resolve();
+            }, 100);
           },
         });
       });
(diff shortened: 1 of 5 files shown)


src/common/Dropdown/DropdownComponent.tsx
```diff
@@ -79,8 +80,10 @@ export const Dropdown = forwardRef<DropdownRef, DropdownProps>(({
             setVisibility(visible ? 'before-appear' : 'before-disappear');
           },
           afterTransition: () => {
-            setVisibility(visible ? 'visible' : null);
-            resolve();
+            setTimeout(() => {
+              setVisibility(visible ? 'visible' : null);
+              resolve();
+            }, 100);
```
bmartel, line 86: What is the significance of 100ms here? How did you come to this number?

juliosgarbi: I'm adding this because here: https://github.com/heartexlabs/label-studio-frontend/blob/14ec33b47560f5720777f4f813a62e7dab9cb725/src/utils/transition.ts#L39 is being added a timeout of 30ms to transition, as this file is an util file and used globally I don't think that is a good idea remove this timeout and this timeout is making the transition doesn't desapear. So I added a 100ms timeout in the afterTransition so every time that the whole process is triggered we are sure that afterTransition will be triggered after transition

bmartel: Hmm, I wonder if this is the right approach. For example, we have this magic number of 30ms in the util, and now we are shimming a use of our own callback to fix the solution. But the original value seems suspect even, what is the significance of these numbers. What is expected to have run inside of these time frames? Can't this just use 

```tsx
setTimeout(() => transition?.(element), 1);
// or even 0?? This seems to just have to make the next render tick in React
setTimeout(() => transition?.(element), 0);
```

and then usages of AfterTransition really should be properly scheduled from the helper to be `after` the initial transition is completed. I think this is possibly just covering a bug.

bmartel (approved):

PR #1354, opened 2023-05-05
feat: LSDV-3025-14: Implement empty state when outliner filter returns empty

juliosgarbi:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

When user is adding a new filter rule on the outliner filter that is returning empty result, the outliner is getting empty

When user is adding a new filter rule on the outliner filter that is returning empty result, the outliner is showing a message that there is result but the filter doesn't match with any.

Screenshot 2023-05-05 at 11 35 17 AM

none

no

no

na

no

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


src/components/SidePanels/OutlinerPanel/OutlinerPanel.tsx
```diff
@@ -69,6 +70,10 @@ const OutlinerStandAlone: FC<OutlinerPanelProps> = ({ regions }) => {
     regions.setFilteredRegions(value);
   }, [regions]);
 
+  useEffect(() => {
+    console.log('regions', regions?.filter);
+  }, [regions?.filter?.length]);
```
bmartel, line 75:
```suggestion
```

src/components/SidePanels/OutlinerPanel/OutlinerPanel.tsx
```diff
@@ -80,7 +85,13 @@ const OutlinerStandAlone: FC<OutlinerPanelProps> = ({ regions }) => {
         onGroupingChange={onGroupingChange}
         onFilterChange={onFilterChange}
       />
-      {regions?.regions?.length > 0 ? (
+      {(regions?.regions?.length > 0 && regions?.filter?.length === 0) ? (
```
bmartel, line 88:
```suggestion
      {(!regions?.regions?.length && regions?.filter?.length) ? (
```
Isn't this a check **if** there are filters but no results?

bmartel: Which property holds the filtered region results, and which holds the active filters?

bmartel (changes requested):

src/components/SidePanels/OutlinerPanel/OutlinerPanel.tsx
juliosgarbi: ops

src/components/SidePanels/OutlinerPanel/OutlinerPanel.tsx
juliosgarbi: It doesn't work regions.filter.length starts as undefined. If we use this validation it starts as true and it shows the message that there are no items to be shown in this filter. And regions.filter needs to be undefined because when its undefined outliner is showing regions.regions otherwise it will show regions.filter.

bmartel (approved):

PR #1360, opened 2023-05-10
feat: LSDV-5033: Add the option to use WebAudioPlayer for Audio

bmartel:

_(check all that apply)_
- [ ] Product design
- [X] Frontend

There was a reported issue which stated the changes to Audio where we moved from using AudioContext and WebAudio API to **play** the audio, to html5 media element to control playback, that the visual drawn **waveform did not align** with the playback auditory cues. Upon further investigation into these particular files, it was found that the mp4 file has an audio track which is truncated, and is much shorter in duration to the video track. This caused a discrepancy seemingly in how the media element reported and played back the audio of this file, and seemed to be out of sync as a result, which equally was out of sync with what was visualized.

As a result we cannot really detect and fix these types of situations, as they are non standard and would have us guessing as to what is the **true position** of the audio track with respect to the video track. (I tried detecting and padding with silence bits, which did work visually, but the underlying media playback was still wrong so it didn't quite work). With that, we are going to offer a configuration to allow users to opt-in to WebAudio as a `player` source. That way **if** a user finds themselves in a scenario which has some oddities presenting like this case (truncated audio track in a video file, or other discrepancies like latency/accuracy to playback), there is a an alternative they can utilize on a project by project basis, and we continue to leave all other cases untouched.

Introduces a new config option to the audio tag:

`<Audio player="html5|webaudio" />`

Defaults to `html5` as is currently the case, and allows a user to opt-in to using WebAudio as the playback controller.

Tradeoffs for these options:

**HTML5**
- Supports chunked buffering for playback (and therefore larger audio durations)
- Uses least amount of required memory
- Not as accurate for fine-grain playback, play/pause/loop and sync can be a bit lagged at times due to buffering mechanisms
- Supports either decoder options of `webaudio` or `ffmpeg` (wasm)

**WebAudio**
- Requires entire file to be put into memory
- Uses more memory as a result of above
- Has much finer grain control over the audio track, and has very low latency ability to play/pause/loop and sync as there is no buffering required.
- Only supports the `webaudio` decoder

We currently only support implicitly the `html5` player option.

N/A

This will impact user's performance, for those who opt into the `webaudio` player option. As stated in the tradeoffs, this option trades accuracy for memory requirements. So this does limit upper bound performance for `webaudio`, dependent of course on the user's machine and operating conditions.

No.

- Padding the visual data with silence through the following calculation:
```
const paddingSamples = new Float32Array(Math.max(0, Math.round((mediaDuration - audioBufferDuration) * sampleRate))))

const visualChunks = [audioBufferSamples, paddingSamples]
```
This worked really well visually, but the underlying problem is that html5 media player is incorrectly skewing the audio track placement (almost as though its time stretching it to match video). So we couldn't entertain this idea any further unfortunately.

Seeing as how this is a direct refactor behind a config option, it should be safe to just test as is with either option enabled. If there is any missing portions, it will be caught in testing.

_(check only one)_
- [X] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

Audio


src/lib/AudioUltra/Controls/Html5Player.ts
```diff
@@ -0,0 +1,82 @@
...
+    this.audio.el.addEventListener('ended', this.handleEnded);
+    this.bufferPromise = new Promise(resolve => {
+      this.bufferResolve = resolve;
+    });
+
+    const time = this.currentTime;
+
+    this.audio.el
+      .play()
+      .then(() => this.bufferPromise!.then())
```
hlomzik, line 38: could this be replaced with this more readable version?
```
Promise.all([
  this.audio.el.play(),
  this.bufferPromise,
])
```

src/lib/AudioUltra/Controls/Html5Player.ts
```diff
@@ -0,0 +1,82 @@
...
+      .then(() => {
+        this.timestamp = performance.now();
+
+        // We need to compensate for the time it took to load the buffer
+        // otherwise the audio will be out of sync of the timer we use to
+        // render updates
+        if (this.audio?.el) {
```
hlomzik, line 45: can it be false at this moment?

src/lib/AudioUltra/Controls/Html5Player.ts
```diff
@@ -0,0 +1,82 @@
...
+          // This must not be notifying of this adjustment otherwise it can cause sync issues and near infinite loops
+          this.setCurrentTime(time);
```
hlomzik, line 47: and don't we need to really compensate this time? player can be playing for some time already or that's impossible?

hlomzik (changes requested):

src/lib/AudioUltra/Controls/Html5Player.ts
bmartel: This is compensating for the time it takes for the buffer to resolve, it captures it before this promise chain.

https://github.com/heartexlabs/label-studio-frontend/pull/1360/files#diff-d26452e66010ef645ec1623dbb83eff35c0a81c26cce13a14f0dd3e2331366d9R34

src/tags/object/AudioUltra/model.js
```diff
@@ -90,6 +90,7 @@ import { WS_SPEED, WS_VOLUME, WS_ZOOM_X } from './constants';
  * @param {string} [waveheight=32] - Minimum height of a waveform when in `splitchannels` mode with multiple channels to display.
  * @param {boolean} [splitchannels=false] - Display multiple audio channels separately, if the audio file has more than one channel. (**NOTE: Requires more memory to operate.**)
  * @param {string} [decoder=webaudio] - Decoder type to use to decode audio data. (`"webaudio"` or `"ffmpeg"`)
+ * @param {string} [player=html5] - Player type to use to play audio data. (`"html5"` or `"webaudio"`)
```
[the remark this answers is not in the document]
bmartel: Which text?

src/lib/AudioUltra/Waveform.ts
```diff
@@ -194,6 +201,11 @@ export class Waveform extends Events<WaveformEventTypes> {
       params.timeline = { placement: 'top' };
     }
 
+    params.decoderType = params.decoderType ?? 'webaudio';
+    // Need to restrict ffmpeg to html5 player as it doesn't support webaudio
+    // because of chunked decoding raw Float32Arrays and no AudioBuffer support
+    params.playerType = params.decoderType === 'ffmpeg' ? 'html5' : params.playerType ?? 'html5';
```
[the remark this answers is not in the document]
bmartel: Definitely will, along with overall tradeoffs.

src/lib/AudioUltra/Controls/Html5Player.ts
bmartel: Yep, can replace that.

src/lib/AudioUltra/Controls/Html5Player.ts
bmartel: The case that comes to mind is hitting play on a large file, and it hits an unbuffered chunk, then the user clicks away on QuickView.

src/lib/AudioUltra/Controls/Html5Player.ts
bmartel: This code is entirely the same as previous.

src/tags/object/AudioUltra/model.js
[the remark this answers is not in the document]
hlomzik: like this one, about possible reasons to use this parameter
> scenario which has some oddities presenting like this case (truncated audio track in a video file, or other discrepancies like latency/accuracy to playback)

hlomzik (approved):

src/lib/AudioUltra/Controls/Html5Player.ts
hlomzik: I was just wondering about possible edge cases, but yeah, you're right, that's not related anyway!

hlomzik:
Approved it, great change! Just some docs pieces are better to be improved.

PR #1477, opened 2023-06-21
fix: LSDV-4937: Label missing from outliner when multiple labels are selected

juliosgarbi:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

When user was selecting two or more labels from different <Labels> for the same annotation the labels weren't being shown in the outliner sidebar, also if user was selecting two or more labels from the labels that has the property `choice="multiple"` set, in the outliner all labels was being shown but all with the same color, normally the same color as the first region selected.

#### What does this fix?
For both cases, now all selected regions are shown in the outliner sidebar and they are being shown with their colors

none

no

no

NA

NA

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [x] integration
- [ ] unit


src/components/SidePanels/OutlinerPanel/OutlinerTree.tsx
```diff
@@ -129,13 +129,33 @@ const useDataTree = ({
...
+        const labels: any[] = [].concat(...labelsInResults);
+
+        return (
+          <Block name="labels-list">
+            {labels.map((label, index) => {
+              const color = label.background || '#000000';
+
+              return [
+                index ? ', ' : null,
+                <span key={label.id} style={{ color }}>
```
yyassi-heartex, line 152: I see we're using a span here - can we use Elem and give it a name instead?

yyassi-heartex: this is optional so i'll go ahead and approve

yyassi-heartex (approved):

src/components/SidePanels/OutlinerPanel/OutlinerTree.tsx
```diff
@@ -129,13 +129,33 @@ const useDataTree = ({
...
         return 'No Label';
       } else if (type.includes('label')) {
         return item.value;
       } else if (type.includes('region') || type.includes('range')) {
-        return (item?.labels ?? []).join(', ') || 'No label';
+        const labelsInResults = item.results
+          .filter((result: any) => result.type.endsWith('labels'))
+          .map((result: any) => result.selectedLabels || []);
+
+        const labels: any[] = [].concat(...labelsInResults);
```
hlomzik, line 143: it's simplier to do this via `[... labelsInResults]`, but you don't need this at all — `labelsInResults` is already a copy

src/components/SidePanels/OutlinerPanel/OutlinerTree.tsx
```diff
@@ -129,13 +129,33 @@ const useDataTree = ({
...
+
     const label = (() => {
       if (!type) {
         return 'No Label';
       } else if (type.includes('label')) {
         return item.value;
       } else if (type.includes('region') || type.includes('range')) {
-        return (item?.labels ?? []).join(', ') || 'No label';
+        const labelsInResults = item.results
+          .filter((result: any) => result.type.endsWith('labels'))
```
hlomzik, line 140: you can type array itself so next array methods don't need to be typed: `(item.results as any)`

hlomzik: also you can use `item.labelings` which return results with labels

src/components/SidePanels/OutlinerPanel/OutlinerTree.tsx
```diff
@@ -129,13 +129,33 @@ const useDataTree = ({
...
+          .map((result: any) => result.selectedLabels || []);
```
hlomzik, line 141: the problem here is that `selectedLabels` can be not anly an array, but a simple label (when related labelset has `allowEmpty`), so it's better to check for it.

src/mixins/DrawingTool.js
```diff
@@ -135,10 +135,13 @@ const DrawingTool = types
...
 
-        const newArea = self.annotation.createResult(value, currentArea.results[0].value.toJSON(), control, obj);
+        const [main, ...rest] = currentArea.results;
+        const newArea = self.annotation.createResult(value, main.value.toJSON(), control, obj);
+
+        rest.forEach(r => newArea.addResult(r.toJSON()));
 
         currentArea.setDrawing(false);
-        self.applyActiveStates(newArea);
+        // self.applyActiveStates(newArea);
```
hlomzik, line 144: remove it

src/mixins/DrawingTool.js
```diff
@@ -135,10 +135,13 @@ const DrawingTool = types
           return value;
         }, { coordstype: 'px', dynamic: self.dynamic });
 
-        const newArea = self.annotation.createResult(value, currentArea.results[0].value.toJSON(), control, obj);
+        const [main, ...rest] = currentArea.results;
+        const newArea = self.annotation.createResult(value, main.value.toJSON(), control, obj);
+
+        rest.forEach(r => newArea.addResult(r.toJSON()));
```
hlomzik, line 141: it's better to add comment why do we do this, all this behaviour is not very obvious and common

hlomzik (changes requested):

src/components/SidePanels/OutlinerPanel/OutlinerTree.tsx
juliosgarbi: I have to do this in this case because sometimes labelsInResult is returning 2 arrays or more arrays with arrays inside with more than 1 node, so here I'm breaking the parent array and concatenating the children arrays in just 1 array, if I do [...labelInResult] the result can be [array(1), array(3)]

src/components/SidePanels/OutlinerPanel/OutlinerTree.tsx
juliosgarbi: done

src/components/SidePanels/OutlinerPanel/OutlinerTree.tsx
juliosgarbi: Can you give me an exemple? I try to reproduce the selectedLabels returning a simple label to understand what I should check, but I wasn't able to. Every test that I made it's returning an array or returning undefined.

src/components/SidePanels/OutlinerPanel/OutlinerTree.tsx
hlomzik: oh, yes, sorry, missed that! so you can use `.flat(1)` for this, a bit more clear than `concat`

hlomzik (approved):

PR #1492, opened 2023-07-05
feat: LSDV-5295: Scroll management of paragraph transcriptions

juliosgarbi:

_(check all that apply)_
- [x] Product design
- [x] Frontend

Now when user hits play on Audio, the paragraphs will scroll automatically following where the audio is.

The paragraphs doesn't follow scrolling automatically.

fflag_feat_front_lsdv_e_278_contextual_scrolling_short

_(check only one)_
- [x] Yes, and covered entirely by feature flag(s)
- [ ] No

_(check all that apply)_
- [x] e2e
- [ ] integration
- [ ] unit


src/tags/object/Paragraphs/HtxParagraphs.js
```diff
@@ -397,6 +413,41 @@ class HtxParagraphsView extends Component {
         return false;
       });
     });
+
+
+    if (isFF(FF_LSDV_E_278) && this.props.item.contextscroll && item.playingId >= 0 && this.lastPlayingId !== item.playingId && this.state.canScroll) {
+      const _start = this.props.item._value[item.playingId].start;
+      const _end = this.props.item._value[item.playingId].end;
```
yyassi-heartex, line 420:
```suggestion
      const playingItem = this.props.item._value[item.playingId];
      const _start = playingItem.start;
      const _end = playingItem.end;
```

src/tags/object/Paragraphs/HtxParagraphs.js
```diff
@@ -397,6 +413,41 @@ class HtxParagraphsView extends Component {
...
+      const _phaseHeight = root.querySelectorAll('div')[item.playingId]?.offsetHeight || 0;
```
yyassi-heartex, line 421: is there anyway you can get a ref to `root.querySelectorAll('div')[item.playingId]` instead of querying it through the DOM? it would be more efficient that navigating the DOM to get this

src/tags/object/Paragraphs/HtxParagraphs.js
```diff
@@ -397,6 +413,41 @@ class HtxParagraphsView extends Component {
...
+      const _duration = this.props.item._value[item.playingId].duration || _end - _start;
+      const _wrapperHeight = root.offsetHeight;
+      const _wrapperOffsetTop = root.querySelectorAll('div')[item.playingId]?.offsetTop - 8; // 8 is the padding between the phrases, so it will keep aligned with the top of the phrase
```
yyassi-heartex, line 424: can we define the 8 as a const variable some where so we can set it once and reuse it as needed

src/tags/object/Paragraphs/HtxParagraphs.js
juliosgarbi: It's being created in another component, it would be possible but not in a correct way I guess.

bmartel: Querying the whole list each time to find the indexed id to pinpoint the active paragraph segment could be quite inefficient, especially depending on how long the list becomes and may also produce unwanted results if this list is ever virtualized or handled differently within React itself. If we were to instead pass an activeRef and set it within the child to point to the current active segment:

```tsx
// child paragraph
ref={isPlaying && activeRef}
```
that would work as well, then you can instead just check:

```tsx
// Paragraphs parent
activeRef.current?.offsetHeight || 0;
```

bmartel (approved):

src/tags/object/Paragraphs/HtxParagraphs.js
```diff
@@ -454,14 +518,34 @@ class HtxParagraphsView extends Component {
...
+        </div>
         <div
           ref={this.myRef}
+          data-testid="phrases-wrapper"
           data-update={item._update}
           className={contextScroll ? styles.scroll_container : styles.container}
           onMouseUp={this.onMouseUp.bind(this)}
         >
-          <Phrases item={item} />
+          <Phrases item={item} playingId={item.playingId} activeRef={this.activeRef} />
```
bmartel, line 548:
```suggestion
          <Phrases item={item} {...(isFF(FF_LSDV_E_278) ? { playingId: item.playingId activeRef: this.activeRef }: {})} />
```

PR #1523, opened 2023-08-07
fix: LSDV-5248: BBoxes optimizations

Gondragos:
Previously, we used the deferred action call approach for double-clicking on a region. It was caused by the fact that there was no way to catch dblclick interaction 'cause re-rendering  elements on selection prevents emiting events.

Right now this problem should be solved by using `Portal` element of `Konva`. As long as the elements are still the same we can catch second click on them so we do not need ta wait anymore.

_(check all that apply)_
- [ ] Product design
- [x] Frontend

N/A

Nope

Nope

`fflag_fix_front_lsdv_5248_double_click_delay_280823_short`

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [x] integration
- [ ] unit

`Image Segmentation`


src/mixins/KonvaRegion.js
```diff
@@ -103,6 +105,16 @@ export const KonvaRegionMixin = types.model({})
 
         if (e) e.cancelBubble = true;
 
+        if (isFF(FF_DBLCLICK_DELAY)) {
+          const isInDoubleClickWindow = lastClick && (ev.timeStamp - lastClick.time < DOUBLE_CLICK_MS_WINDOW);
+          const isSamePosClick = lastClick && (Math.abs(ev.offsetX - lastClick.x) < MAX_CLICK_POS_DEVIATION && Math.abs(ev.offsetY - lastClick.y) < MAX_CLICK_POS_DEVIATION);
+
+          if (isInDoubleClickWindow && isSamePosClick) {
```
yyassi-heartex, line 112: this is a cool approach but involves checking time stamps to confirm if double click happened or not - is there a reason we didnt use `ev.detail`? if `ev.detail = 1` then its 1 click, `ev.detail = 2` means double click

Gondragos: Wow. Somehow I literally didn't know about that detail property. Thanks a lot! It makes my life easier.

yyassi-heartex (approved):

src/regions/AliveRegion.tsx
```diff
@@ -28,13 +38,20 @@ export const AliveRegion = (
 
   return observer(({ item, ...rest }: RegionComponentProps) => {
     const canRender = options?.renderHidden || !item.hidden;
+    const shouldNotUsePortal = isFF(FF_DBLCLICK_DELAY) || options?.shouldNotUsePortal;
```
hlomzik, line 41: `options.shouldNotUsePortal` is set to true only in one region (Brush) and the value is `isFF(FF_DBLCLICK_DELAY)`, so it can be just removed.

src/regions/AliveRegion.tsx
```diff
@@ -28,13 +38,20 @@ export const AliveRegion = (
...
+    const Wrapper = (shouldNotUsePortal ? Fragment : Portal) as ExoticComponent<PortalProps>;
```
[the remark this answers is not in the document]
hlomzik: and with FF on we basically have a Fragment, which results in no tags at all, so all this code can be removed?

hlomzik (changes requested):

src/components/ImageView/ImageView.js
```diff
@@ -388,7 +392,10 @@ const SelectionLayer = observer(({ item, selectionArea }) => {
 const Selection = observer(({ item, selectionArea, ...triggeredOnResize }) => {
   return (
     <>
-      <SelectedRegions item={item} selectedRegions={item.selectedRegions} {...triggeredOnResize} />
+      { !isFF(FF_DBLCLICK_DELAY)
```
hlomzik, line 395: reverse condition for readability, please

src/components/ImageView/ImageView.js
```diff
@@ -574,7 +582,18 @@ export default observer(
           e.target === item.stageRef ||
           findClosestParent(
             e.target,
-            el => el.nodeType === 'Group' && ['ruler', 'segmentation'].indexOf(el?.attrs?.name) > -1,
+            el => {
+              if (el.nodeType === 'Group') {
+                if ('ruler' === el?.attrs?.name) {
+                  return true;
+                }
+                if ((!isFF(FF_DBLCLICK_DELAY) || !isMoveTool)
```
hlomzik, line 590: please, rewrite this! extract lambda into separate named function, add comment, simplify condition

src/mixins/KonvaRegion.js
```diff
@@ -35,7 +34,7 @@ export const KonvaRegionMixin = types.model({})
     };
   })
   .actions(self => {
-    let deferredSelectId = null;
+    let deferredSelectId;
```
hlomzik, line 37: are there some importance in this change from `null` to `undefined`?

src/mixins/KonvaRegion.js
```diff
@@ -113,21 +121,29 @@ export const KonvaRegionMixin = types.model({})
...
+              annotation.selectArea(self);
+            } else {
+              deferredSelectId = setTimeout(selectAction, 300);
+            }
           }
         }
       },
+      onDoubleClickRegion() {
+        self.requestPerRegionFocus();
+        self.annotation.selectAreas([self]);
```
hlomzik, line 146: add comment here why can't we use usual `selectArea()`

hlomzik (changes requested):

src/mixins/KonvaRegion.js
Gondragos: Nope, it's just an accident

PR #1535, opened 2023-08-22
fix: LSDV-5526: Fix duplication of classifications in LLM scenario

Gondragos:
This PR should fix some issues in existed LLM scenario such as:
- After editing text and with a successful response, textarea doesn't get replaced (it also causes areas duplication)
- Getting errors from ml-backend breaks the current regions state

_(check all that apply)_
- [ ] Product design
- [x] Frontend

#### What does this fix?
It fixes areas duplication and stop from making any changes in region and suggestion lists if we get null instead of new suggestion list. 

N/A

Nope

Nope

The alternative approach could be provided in the context of another tasks when we get understanding of what it should be.

`fflag_feat_all_lsdv_e_294_llm_annotations_180723_long`

_(check only one)_
- [ ] No

Enabling `flag_feat_all_lsdv_e_294_llm_annotations_180723_long` could break the previous scenarios used smart tools for Image Segmentation. 

_(check all that apply)_
- [ ] e2e
- [x] integration
- [ ] unit

`Auto Annotating`, `TextArea`


src/stores/Annotation/Annotation.js
```diff
@@ -1274,12 +1275,40 @@ export const Annotation = types
 
     acceptSuggestion(id) {
       const item = self.suggestions.get(id);
+      let itemId = id;
+      const isGlobalClassification = item.classification;
+
+      if (isFF(FF_LLM_EPIC)) {
```
hlomzik, line 1281: add a comment, please, that this piece of code prevents from creating duplicated global classifications

tests/functional/specs/llm/accept_suggestions.cy.ts
```diff
@@ -0,0 +1,54 @@
...
+    const promptTextAarea = useTextarea('&:eq(0)');
+    const answerTextArea = useTextarea('&:eq(1)');
+
+    promptTextAarea.type(`${LLM_WHAT_DO_YOU_SEE.prompt}{shift+enter}`);
+    answerTextArea.hasValue(LLM_WHAT_DO_YOU_SEE.answer);
+    promptTextAarea.clickRowEdit(1);
+    promptTextAarea.rowType(1, `{selectAll}{backspace}${LLM_WHAT_ELSE_DO_YOU_SEE.prompt}{shift+enter}`);
+    answerTextArea.hasValue(LLM_WHAT_ELSE_DO_YOU_SEE.answer);
+
+    LabelStudio.serialize().then((result) => {
```
hlomzik, line 50: should not it be async for proper test finishing detection?

tests/functional/specs/llm/accept_suggestions.cy.ts
```diff
@@ -0,0 +1,54 @@
...
+  const prompt = area?.results?.[0]?.value?.text?.[0];
+
+  if (prompt) {
+    const suggestions = llmTextareaSuggestions[prompt];
+
+    setTimeout(() => {
+      area.store.loadSuggestions(new Promise((resolve) => setTimeout(() => {
+        resolve(suggestions);
+      }, 1000)), x => x);
+    }, 1000);
```
hlomzik, line 20: why so long? can we reduce it to make it go faster?

hlomzik (approved):

tests/functional/specs/llm/accept_suggestions.cy.ts
Gondragos: We surely can, but I don't know if it catch the problem when it appears.

tests/functional/specs/llm/accept_suggestions.cy.ts
Gondragos: It looks like it should not. It waits for resolve as I see https://docs.cypress.io/api/commands/then#Yields. And it looks like Cypress is aware of these processes and also waits for it.

src/stores/Annotation/Annotation.js
```diff
@@ -1290,8 +1319,10 @@ export const Annotation = types
       // hack to unlock sending textarea results
       // to the ML backen every time
       // it just sets `fromSuggestion` back to `false`
-      const isTextArea = area.results.findIndex(r => r.type === 'textarea') >= 0; 
+      const isTextArea = area.results.findIndex(r => r.type === 'textarea') >= 0;
 
+      // This is temporary exception until we find the way to do it right
+      // and this was done to keep notifications on prompt editing or fixing answer from ML backend
```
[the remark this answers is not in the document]
Gondragos: We have a couple of tickets about LLM. And in one of them we could rethink this approach. But it isn't guaranteed.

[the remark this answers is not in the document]
yyassi-heartex: ok, i think thats enough to ensure it doesnt fall through the cracks - which concerned about - but if there is a plan to look at this again as part the LLM tickets we should be ok

yyassi-heartex (approved):

PR #1552, opened 2023-08-31
feat: OPTIC-2: Save draft on leave

Travis1282:

_(check all that apply)_
- [X] Product design
- [X] Frontend

lost drafts during timeout, no user feedback

#### What does this fix?
_(if this is a bug fix)_

saved drafts when user navigates away from annotation page, gives toast feedback

lost drafts during timeout, no user feedback

_(check only one)_
- [X] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


src/stores/Annotation/Annotation.js
```diff
@@ -677,14 +677,22 @@ export const Annotation = types
       self.setDraftSelected();
       self.versions.draft = result;
       self.setDraftSaving(true);
-
-      return self.store.submitDraft(self, params).then(self.onDraftSaved);
+      return self.store.submitDraft(self, params).then((res) => {
+        self.onDraftSaved;
```
bmartel, line 681:
```suggestion
  self.onDraftSaved(res);
```
This was previously getting invoked but now is not.

bmartel (approved):

src/stores/Annotation/Annotation.js
```diff
@@ -677,14 +677,22 @@ export const Annotation = types
       self.setDraftSelected();
       self.versions.draft = result;
       self.setDraftSaving(true);
-
-      return self.store.submitDraft(self, params).then(self.onDraftSaved);
+      return self.store.submitDraft(self, params).then((res) => {
+        self.onDraftSaved(res);
```
bmartel, line 681: Does any other callers of saveDraft require that the onDraftSaved callback be the return of the function promise? I noticed we are changing to use the raw result instead so that we can get access to this information, I'm wondering if that might break other usages. Do we know for certain the usages expectation of this return?

bmartel: I'm going to approve this with the caveat that we may have to consider this depending on whether any integration tests fail (and subsequently answer this question for us 😅).

PR #1559, opened 2023-09-13
fix: LSDV-5547: Support HTML in Instructions modal

hlomzik:
Format instructions as HTML the same way we do it in LS/LSE

- [ ] Product design
- [x] Frontend

Discrepancy between Label Stream behavior, where instructions displayed in intro modal, formatted as HTML, and instructions modal opened by (i) button, where all HTML displayed as plain text.

#### What does this fix?
Formats instruction as HTML

It dangerously displays instructions as HTML, this can use any scripts etc. But the same is used already in LS/LSE. And this instruction can be set only by managers.

This exact task made as it should be. But we can add some layer of security, sandboxing the modal.

- [x] No

- [ ] e2e
- [ ] integration
- [ ] unit

Instructions, Label Stream

### DIFF
--- a/src/components/InstructionsModal/InstructionsModal.tsx
+++ b/src/components/InstructionsModal/InstructionsModal.tsx
@@ -8,15 +8,17 @@ export const InstructionsModal = ({
   onCancel,
 }: {
   title: string,
-  children: string,
+  children: React.ReactNode,
   visible: boolean,
   onCancel: () => void,
 }) => {
+  const contentStyle: Record<string, string> = { padding: '0 24px 24px', whiteSpace: 'pre-wrap' };
+
   return (
     <>
       <Modal
         title=""
-        visible={visible}
+        open={visible}
         maskClosable
         footer={null}
         closable={true}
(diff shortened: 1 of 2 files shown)


src/components/InstructionsModal/InstructionsModal.tsx
```diff
@@ -8,7 +8,7 @@ export const InstructionsModal = ({
   onCancel,
 }: {
   title: string,
-  children: React.ReactNode,
+  children: string,
```
yyassi-heartex, line 11: cant it it be both?
```suggestion
  children: React.ReactNode|string,
```

src/components/InstructionsModal/InstructionsModal.tsx
```diff
@@ -45,7 +45,10 @@ export const InstructionsModal = ({
         >
           {title}
         </h2>
-        <p style={{ padding: '0 24px 24px', whiteSpace: 'pre-wrap' }}>{children}</p>
+        <p
+          style={{ padding: '0 24px 24px', whiteSpace: 'pre-wrap' }}
+          dangerouslySetInnerHTML={{ __html: children }}
```
yyassi-heartex, line 50: if we're considering using `dangerouslySetInnerHTML` on children which i think comes from user input - should we consider pulling in https://github.com/cure53/DOMPurify to sanitize it a little?

src/components/InstructionsModal/InstructionsModal.tsx
hlomzik: This will go directly to `__html` and I can't see how react nodes can be converted there. We can use them directly though, but that's definitely not the case now. So I'd leave it, because this type reflects both utilization and invocation of this param.

src/components/InstructionsModal/InstructionsModal.tsx
hlomzik: It's actually already safe because of backend sanitization!
https://github.com/HumanSignal/label-studio/blob/eece1c8d39aa3850829bbc9b9b74833b2c5fba26/label_studio/projects/serializers.py#L88-L91

yyassi-heartex (approved):
cool, makes sense - approved

PR #1561, opened 2023-09-14
fix: LSDV-5560: Fix plus and minus hotkeys

Gondragos:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

`ctrl+plus` and `ctrl-minus` hotkeys do not work for zoom in Image Segmentation.

#### What does this fix?
It adds aliases for `plus` and `minus` comands to make it work with original keymaster.

`@heartexlabs/ls-test`

Nope

Nope

Forking and changing `keymaster`

N/A

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [x] integration
- [ ] unit

`Hotkeys`, `Image Tools`

### DIFF
--- a/src/core/Hotkey.ts
+++ b/src/core/Hotkey.ts
@@ -167,10 +167,7 @@ export const Hotkey = (
     applyAliases(key: string) {
       return key
         .split(',')
-        .map(k => {
-          return k.split('+').map(k => ALIASES[k.trim()] ?? k).join('+');
-        },
-        )
+        .map(k => k.split('+').map(k => ALIASES[k.trim()] ?? k).join('+'))
         .join(',');
     },
     /**
--- a/src/core/Hotkey.ts
+++ b/src/core/Hotkey.ts
@@ -80,7 +80,7 @@ keymaster.filter = function(event) {
 };
 
 const ALIASES = {
-  'plus': '=',
+  'plus': '=', // "ctrl plus" is actually a "ctrl =" because shift is not used
   'minus': '-',
 };
 
--- a/src/core/Hotkey.ts
+++ b/src/core/Hotkey.ts
@@ -79,6 +79,11 @@ keymaster.filter = function(event) {
   return true;
 };
 
+const ALIASES = {
+  'plus': '=',
+  'minus': '-',
+};
+
 export const Hotkey = (
   namespace = 'global',
   description = 'Hotkeys',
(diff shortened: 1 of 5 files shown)


src/core/Hotkey.ts
```diff
@@ -79,6 +79,11 @@ keymaster.filter = function(event) {
   return true;
 };
 
+const ALIASES = {
+  'plus': '=',
```
yyassi-heartex, line 83: plus is '='? is that intentional? seems a little misleading

Gondragos: It's the same button. And there is no way we can make it real "+" using `keymaster` library.

yyassi-heartex (approved):

src/core/Hotkey.ts
```diff
@@ -159,6 +164,15 @@ export const Hotkey = (
   _destructors.push(unbind);
 
   return {
+    applyAliases(key: string) {
+      return key
+        .split(',')
+        .map(k => {
+          return k.split('+').map(k => ALIASES[k.trim()] ?? k).join('+');
+        },
+        )
```
hlomzik, line 173:
```suggestion
        .map(k => {
          return k.split('+').map(k => ALIASES[k.trim()] ?? k).join('+');
        })
```
or even
```suggestion
        .map(k => k.split('+').map(k => ALIASES[k.trim()] ?? k).join('+'))
```

tests/functional/yarn.lock
```diff
@@ -264,6 +264,28 @@
   resolved "https://registry.yarnpkg.com/@discoveryjs/json-ext/-/json-ext-0.5.7.tgz#1d572bfbbe14b7704e0ba0f39b74815b84870d70"
   integrity sha512-dBVuXR082gk3jsFp7Rd/JI4kytwGHecnCoTtXFb7DB6CNHp4rg5k1bhg0nWdLGLnOV71lmDzGQaLMy8iPLY0pw==
 
+"@heartexlabs/ls-test@git+ssh://git@github.com/heartexlabs/ls-frontend-test#36fc7ffc2cddc4e77aa785ec3304900b27163f72":
```
hlomzik, line 267: replace the next occurency, I don't know why did it left it intact

hlomzik (approved):

tests/functional/yarn.lock
Gondragos: Somehow it works fine with that. I mean sometimes `yarn add` keeps it and sometimes removes it but it still works so I don't think we need to remove it manually every time.

hlomzik: I just wanted to highlight that if yarn puts it here most probably it's because of some local leftovers. But it really affects nothing, so okay

PR #1581, opened 2023-10-09
fix: LEAP-235: Fix the property showFullPath to work as it was working in the old taxonomy

juliosgarbi:
- [ x Commit message(s) and PR title follows the format `[fix|feat|ci|chore|doc]: TICKET-ID: Short description of change made` ex. `fix: DEV-XXXX: Removed inconsistent code usage causing intermittent errors`

_(check all that apply)_
- [ ] Product design
- [x] Frontend

The property showFullPath is not working on the new taxonomy and it should be working as it was working in the old Taxonomy

#### What does this fix?
It is fixing the property showFullPath in the new Taxonomy, with this fix if user set the property showFullPath in the Taxonomy tag, when user selects a item, it will start showing all the path using ' / ' to separate each tag.

No

No

No

As it is a specific fix, I don't think that it would be possible to use an alternative approache

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

### DIFF
--- a/src/components/NewTaxonomy/NewTaxonomy.tsx
+++ b/src/components/NewTaxonomy/NewTaxonomy.tsx
@@ -94,6 +94,21 @@ const NewTaxonomy = ({
     return flatten;
   }, [items]);
 
+  const formatOutputValue = useMemo(() => {
+    return selected.map(path => {
+      const selectedItem = path.map(value => {
+        const label = flatten.find(taxonomyItem => taxonomyItem.path.at(-1) === value)?.label;
+
+        return label ?? value;
+      });
+
+      return {
+        label: options.showFullPath ? selectedItem.join(separator) : selectedItem[path.length - 1],
+        value: path.join(separator),
+      };
+    });
+  }, [selected]);
+
   const loadData = useCallback(async (node: any) => {
     return onLoadData?.(node.value.split(separator));
   }, []);
@@ -102,18 +117,7 @@ const NewTaxonomy = ({
     <TreeSelect
       treeData={treeData}
       labelInValue={true}
-      value={selected.map(path => {
-        const selectedItem = path.map(value => {
-          const label = flatten.find(taxonomyItem => taxonomyItem.path[taxonomyItem.path.length - 1] === value)?.label;
-
-          return label ?? value;
-        });
-
-        return {
-          label: options.showFullPath ? selectedItem.join(separator) : selectedItem[path.length - 1],
-          value: path.join(separator),
-        };
-      })}
+      value={formatOutputValue}
       onChange={items => onChange(null, items.map(item => item.value.split(separator)))}
       loadData={loadData}
       treeCheckable
--- a/src/components/NewTaxonomy/NewTaxonomy.tsx
+++ b/src/components/NewTaxonomy/NewTaxonomy.tsx
@@ -48,19 +48,18 @@ type TaxonomyProps = {
 };
 
 const convert = (items: TaxonomyItem[], options: TaxonomyOptions): AntTaxonomyItem[] => {
-  return items.map(item => {
-    return {
-      title: item.hint ? (
-        <Tooltip title={item.hint} mouseEnterDelay={500}>
-          <span>{item.label}</span>
-        </Tooltip>
-      ) : item.label,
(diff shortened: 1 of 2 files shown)


src/components/NewTaxonomy/NewTaxonomy.tsx
```diff
@@ -48,18 +48,19 @@ type TaxonomyProps = {
...
-        <span>{item.label}</span>
-      </Tooltip>
-    ) : item.label,
-    value: item.path.join(options.pathSeparator),
-    key: item.path.join(options.pathSeparator),
-    isLeaf: item.isLeaf !== false && !item.children,
-    disableCheckbox: options.leafsOnly && (item.isLeaf === false || !!item.children),
-    children: item.children ? convert(item.children, options) : undefined,
-  }));
+  return items.map(item => {
```
hlomzik, line 51: please, revert it back, especially this weird "};});" 😄

src/components/NewTaxonomy/NewTaxonomy.tsx
```diff
@@ -83,15 +84,40 @@ const NewTaxonomy = ({
...
+
+          return label ?? value;
+        });
+
+        return {
+          label: options.showFullPath ? selectedItem.join(separator) : selectedItem[path.length - 1],
+          value: path.join(separator),
+        };
+      })}
+      onChange={items => {
```
hlomzik, line 118: revert `onChange()` as well, please

src/components/NewTaxonomy/NewTaxonomy.tsx
```diff
@@ -83,15 +84,40 @@ const NewTaxonomy = ({
...
 
   return (
     <TreeSelect
       treeData={treeData}
-      value={selected.map(path => path.join(separator))}
-      onChange={items => onChange(null, items.map(item => item.value.split(separator)))}
+      labelInValue={true}
+      value={selected.map(path => {
+        const selectedItem = path.map(value => {
+          const label = flatten.find(taxonomyItem => taxonomyItem.path[taxonomyItem.path.length - 1] === value)?.label;
```
hlomzik, line 108: use `.at()` to get the latest item

src/components/NewTaxonomy/NewTaxonomy.tsx
```diff
@@ -83,15 +84,40 @@ const NewTaxonomy = ({
...
     return onLoadData?.(node.value.split(separator));
   }, []);
 
   return (
     <TreeSelect
       treeData={treeData}
-      value={selected.map(path => path.join(separator))}
-      onChange={items => onChange(null, items.map(item => item.value.split(separator)))}
+      labelInValue={true}
+      value={selected.map(path => {
```
hlomzik, line 106: better to move this at least to callback function here or maybe even to a method in a model. and then flatten can become a getter in a model, which will increase performance and readability a little.

hlomzik (changes requested):

src/components/NewTaxonomy/NewTaxonomy.tsx
```diff
@@ -83,14 +83,41 @@ const NewTaxonomy = ({
...
+    };
+
+    items.forEach(visitItem);
+    return flatten;
+  }, [items]);
+
+  const formatOutputValue = useMemo(() => {
+    return selected.map(path => {
+      const selectedItem = path.map(value => {
+        const label = flatten.find(taxonomyItem => taxonomyItem.path.at(-1) === value)?.label;
```
hlomzik, line 100: By searching for the alias which is not unique through the whole list of items, we can accidentially grab the wrong one. For example with simple ordered numbers as aliases on every level, so the path will be like `["1", "1", "4", "1"]`, we'll get the same item three times, two of them will be the wrong ones.
We should go down the tree as far as we can to find the correct item. This will improve performance (less memory because of dropping `flatten` and fewer passes because of scanning only one branch for every level) and will make it correct. For not loaded items we'll just return the rest of the path.

hlomzik (changes requested):

hlomzik:
closed in favor of https://github.com/HumanSignal/label-studio-frontend/pull/1595

PR #1586, opened 2023-10-11
fix: LEAP-255: Parse username and password from basic auth URL and use for API calls

juliosgarbi:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

User must be able to use the prop apiUrl with basic authorization.

#### What does this fix?
Add the possibility to user use the prop apiUrl as `https://username:password@rest.of.url/` and taxonomy will parse it to authenticate

no

no

no

no

_(check only one)_
- [ ] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

### DIFF
--- a/src/tags/control/Taxonomy/Taxonomy.js
+++ b/src/tags/control/Taxonomy/Taxonomy.js
@@ -351,7 +351,6 @@ const Model = types
 
       try {
         const res = yield fetch(url, requestOptions);
-
         const dataRaw = yield res.json();
         // @todo temporary to support deprecated API response format (just array, no items)
         const data = dataRaw.items ?? dataRaw;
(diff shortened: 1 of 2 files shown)


src/tags/control/Taxonomy/Taxonomy.js
```diff
@@ -337,8 +338,17 @@ const Model = types
...
+        requestOptions = {
+          headers: new Headers({
+            'Authorization': `Basic ${btoa(`${url.username}:${url.password}`)}`,
+          }),
+        };
+      }
+
       try {
-        const res = yield fetch(url);
+        const res = yield fetch(url.origin + url.pathname, requestOptions);
```
jombooth, line 350: :wave: I was just skimming this PR, won't using only origin and pathname cause query params to be dropped? I think instead it'd be safer to set `url.username` and `url.password` to null after moving them onto the authorization header, then continue to yield `fetch(url)`

hlomzik: yes, good catch and suggestion!

hlomzik (changes requested):

hlomzik (approved):

bmartel (approved):

PR #1601, opened 2023-10-23
fix: LEAP-218: Improve performance of search

juliosgarbi:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

The original search tool from `antd tree-select` was having problem with performance when it was being used with a long list of choices, as the list was being highlighted, a lot of elements was being changed at the same time creating a big performance problem

#### What does this fix?
We decided to stop using the original search tool and create our own search tool and attach to tree-select, with our own search tool we have more flexibility and we was able to fix the performance problem removing the highlight when the search matches with the choice item.

none

As the search tool isn't highlighting the matches values, the performance is way better. Now we are using our own search tool to filter the results so we have more flexibility to treat the result as we want

no

We have spent some time trying different approaches, but unfortunately the component `tree-select` from `antd` doesn't give us a lot of options, we were able to change the way that the search was being done, but it's impossible to deactivate the highlight. Even though that would be possible to improve the performance changing the way that the search works, the items being highlighted would keep being a problem to the performance.

_(check only one)_
- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit


bmartel (approved):

src/components/NewTaxonomy/TaxonomySearch.tsx
```diff
@@ -0,0 +1,102 @@
...
+        }
+        return total;
+      }, []);
+    };
+
+    return dig(treeData);
+  }, []);
+
+  // After filter the treeData result, this method is used to check which keys must be expanded
+  const findExpandedKeys = useCallback((data: AntTaxonomyItem[]) => {
```
bmartel, line 59: This could be done within the same tree filter in a single pass for even better performance on larger trees. Just a suggestion, though for now this will be a great improvement over the previous.

src/components/NewTaxonomy/NewTaxonomy.tsx
```diff
@@ -133,14 +139,57 @@ const NewTaxonomy = ({
...
+    setExpandedKeys(expandedKeys);
+  }, []);
+
+  const renderDropdown = useCallback((origin: ReactNode) => {
+    if (!defaultSearch) {
+      return (
+        <TaxonomySearch
+          ref={refInput}
+          treeData={treeData}
+          origin={origin}
```
hlomzik, line 153: `origin` can be moved out of the TaxonomySearch because it's not used there. will make code cleaner there and simplier here

src/components/NewTaxonomy/TaxonomySearch.tsx
```diff
@@ -0,0 +1,132 @@
...
+  useEffect(() => {
+    const _filteredData = filterTreeData(treeData, inputValue);
+
+    onChange(_filteredData.filteredDataTree, _filteredData.expandedKeys);
+  }, [treeData]);
+
+  //To filter the treeData items that match with the searchValue
+  const filterTreeNode = useCallback((searchValue: string, treeNode: AntTaxonomyItem) => {
+    const lowerSearchValue = String(searchValue).toLowerCase();
+    const lowerResultValue = typeof treeNode.title === 'object' ? treeNode.title.props.children.props.children : treeNode.title;
```
hlomzik, line 57: oh what? can you add a comment about this chaining?

src/components/NewTaxonomy/TaxonomySearch.tsx
```diff
@@ -0,0 +1,133 @@
...
+        const childList = dig(children || [], match);
+
+        if (match || childList.length) {
+          if (!keepAll)
+            _expandedKeys.push(dataNode.key);
+
+          total.push({
+            ...dataNode,
+            isLeaf: undefined,
+            children: childList,
```
hlomzik, line 91: `childList` is always an array, which makes it look always expandable even for leaf nodes. more correct will be to call `dig()` only if `children` is array and has items, thus for leaf nodes `childNodes` will be undefined and they won't have expand icons.

hlomzik (changes requested):

src/components/NewTaxonomy/TaxonomySearch.tsx
```diff
@@ -0,0 +1,131 @@
...
+    <Block
+      ref={inputRef}
+      value={inputValue}
+      tag={'input'}
+      onChange={(e: ChangeEvent<HTMLInputElement>) => {
+        setInputValue(e.target.value);
+        handleSearch(e);
+      }}
+      onKeyDown={(e: KeyboardEvent<HTMLInputElement>) => {
+        if (e.key === 'Backspace' || e.key === 'Delete') e.stopPropagation();
```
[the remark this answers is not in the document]
juliosgarbi: Without this when you hit backspace or delete to clear the input it deleting the last selected item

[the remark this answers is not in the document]
hlomzik: oh, gotcha, thanks! and it's better to add a comment as well
```suggestion
        // to prevent selected items from being deleted
        if (e.key === 'Backspace' || e.key === 'Delete') e.stopPropagation();
```

hlomzik (approved):

PR #1605, opened 2023-10-25
fix: LEAP-31: Fix functioning of auto annotations

Gondragos:

_(check all that apply)_
- [x] Product design
- [x] Frontend

Working with data annotations was too complicated and inconsistent in a lot of cases so this PR is trying to change that by fixing bugs and implementing a different approach for the tags that do not support suggestion by design.

#### What does this fix?
- remove FF that could break auto annotating for Image Segmentstion etc.
- fix problem with auto annotation toggle which did not affect previously created regions so we had unexpected requests for sugestions
- fix potential problem with groupping regions by type (getConnectedDynamicRegions)
  - it didn't work with classifications except textarea
  - it didn't work with Audio and Video
  - it had unexpected regions in this list before
- allow classifications, audio and video regions to trigger `regionFinishedDrawing` event (they don't have to be dynamic). And it will always accept suggestions automaticaly.
- fix side effects where some regions not from the region list could generate events for requesting suggestions (it also could generate accidental empty request)
  - it was genereted from accepted suggestions wen they were deleted (accepted)
  - it was generated from TextArea, because to display data correctly it creates quasi-region, using real region model.
- clean/rework related hacks in code

`@heartexlabs/ls-test`

Nope

Nope

Was not detected

N/A

_(check only one)_
- [ ] No

It fixes a lot and it didn't work properly before, but it should be reviewed and tested with full attention.

_(check all that apply)_
- [ ] e2e
- [x] integration
- [ ] unit

`Auto Annotations`, `Smart tools`, `Dynamic regions`, `LLM`, `Image`, `Textarea`


src/stores/Annotation/Annotation.js
```diff
@@ -1337,14 +1341,6 @@ export const Annotation = types
...
       self.suggestions.delete(id);
       
-      // hack to unlock sending textarea results
-      // to the ML backen every time
-      // it just sets `fromSuggestion` back to `false`
-      const isTextArea = area.results.findIndex(r => r.type === 'textarea') >= 0;
-
-      // This is temporary exception until we find the way to do it right
-      // and this was done to keep notifications on prompt editing or fixing answer from ML backend
-      if (isTextArea) area.revokeSuggestion();
```
hlomzik, line 1347: `Region#revokeSuggestion()` can be removed then

hlomzik (approved):

src/mixins/Regions.js
```diff
@@ -86,12 +86,39 @@ const RegionsMixin = types
...
+      if (!self.isRealRegion) return false;
+      if (self.annotation.isSuggestionsAccepting) return false;
+      // There are two modes:
+      // If object tag support suggestions - the region should be marked as a dynamic one to make notifications
+      // If object tag doesn't support suggestions - every region works as dynamic with auto suggestions
+      const canBeReasonOfNotification = self.supportSuggestions ? self.dynamic && !self.fromSuggestion : true;
+
+      const isSmartEnabled = self.results.some(r => {
+        return r.from_name.smartEnabled;
+      });
```
yyassi-heartex, line 118:
```suggestion
      const isSmartEnabled = self.results.some(r => r.from_name.smartEnabled);
```
just to make it a little shorter, its a 1 liner

yyassi-heartex (approved):
looks good to me, i just added 1 little suggestion for readability since its a one liner but thats optional since its fine as it is

src/mixins/Regions.js
```diff
@@ -86,12 +86,37 @@ const RegionsMixin = types
 
       const result = regions.filter(region => {
         if (excludeSelf && region === self) return false;
-        return region.dynamic && region.type === type && region.labelName === labelName;
+        const canBePartOfNotification = self.supportSuggestions ? self.dynamic : true;
+
+        return canBePartOfNotification
+          && region.type === type
+          && region.labelName === labelName
+          && region.results?.[0]?.to_name === self.results?.[0]?.to_name;
```
nicholasrq, line 94: It potentially compare `undefined === undefined` here. Could it lead to any unexpected issues?

src/mixins/Regions.js
```diff
@@ -86,12 +86,37 @@ const RegionsMixin = types
...
       });
 
       return result;
     },
 
+    // Indicates that it is not temporary region created just to display data like Textarea's one
+    // and is not a suggestion
+    get isRealRegion() {
+      return self.annotation?.areas?.has(self.id);
```
nicholasrq, line 103: That's actually smart. Love it.

src/stores/Annotation/Annotation.js
```diff
@@ -252,6 +252,7 @@ export const Annotation = types
     draftSelected: false,
     autosaveDelay: 5000,
     isDraftSaving: false,
+    isSuggestionsAccepting: false,
```
nicholasrq, line 255: could you please add a comment describing this flag? from the name alone it's not very clear what it supposed to do

nicholasrq (approved):

src/mixins/Regions.js
Gondragos: I've checked that. It seems that the only way to get `undefined === undefined` is collecting connected regions in the case with config contains two classification tags with the same `to_name`.  I'm not sure if wee need to group them but it's anyhow is better than it was.

PR #1612, opened 2023-11-07
feat: LEAP-336: Collapsible cards in Ranker

hlomzik:
Allow to collapse and expand cards for easy interactions. With collapsed cards user can see lot of items together with makes rearrangement easier.

Collapsed/expanded states are indicated by ASCII triangles (△/▽ and ▲/▼ for hovered state) used in css to change only class names, not the content and to keep indicators in one place.

Collapse is desirable action so column is treated as expanded when at least one item is expanded. This allows to collapse all items in any state by one click.

New hidden parameter added to `Ranker`: `collapsible` with default true to allow to disable this functionality if it's not needed for any reason.

none

nope

nope

Cards can be made hidden with animation to not change the content inside, but this will require calculations of their height.

none yet

- [x] No

_(check all that apply)_
- [ ] e2e
- [ ] integration
- [ ] unit

Ranker

Screenshot 2023-11-07 at 16 14 40

### DIFF
--- a/src/components/Ranker/Ranker.module.scss
+++ b/src/components/Ranker/Ranker.module.scss
@@ -40,10 +40,10 @@
       cursor: pointer;
     }
 
-    button span::after {
+    &.expanded button span::after {
       content: '△';
     }
-    button:hover span::after {
+    &.expanded button:hover span::after {
       content: '▲';
     }
     &.collapsed button span::after {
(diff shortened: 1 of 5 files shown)


src/components/Ranker/Ranker.module.scss
```diff
@@ -27,6 +27,32 @@
...
+      margin-left: auto;
+      background: none;
+      border: none;
+      font-size: 0.75em;
+      width: 2em;
+      cursor: pointer;
+    }
+
+    button span::after {
+      content: '△';
```
yyassi-heartex, line 44: this functions just fine but it may look better if we use icons instead - can Design provide icons? or do you want to pull in React icons much like we have in DM2 and LSP

yyassi-heartex (approved):
approved but i have a optional suggestion to use icons like react-icons or icons provided by design for better visuals

PR #1613, opened 2023-11-08
ci: New PR template

hlomzik:
Shorter version of our frontend PR template to encourage and help to make more thoughtful description and highlight important nuances of every PR.

Current PR template usually left (mostly) untouched, because it has lot of items to read and change.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
 Main description of what's done here, shortly, with reasoning, references, thoughts and ideas for better understanding.
-This text is intended to be removed and replaced with actual description. Other items can also be removed or edited if not relevant. If any of following items require more description, it should also go in this block.
-Please, read all items and make an action every time to create more conscious PR. Thanks!
+This text is intended to be replaced with actual description. Other items should be filled in or removed if not relevant. If any of following items require more description, it should also go in this block.
+Please read all items and ensure each section has been **filled in or removed** to create more consistent pull requests! Thanks!
 
 ### PR fulfills these requirements
 - [ ] Tests for the changes have been added/updated
@@ -9,11 +9,13 @@ Please, read all items and make an action every time to create more conscious PR
 - [ ] Self-reviewed and ran all changes on a local instance
 
 ### Describe the reason for change
-(if it's not clear from description)
+_(if it's not clear from description — otherwise remove section)_
 
 ### What feature flags were used to cover this change?
+_(feature flags created for this pull request or used to cover changes in this pull request)_
 
 ### What alternative approaches were there?
+_(if there are multiple ways to go, describe them and the reasoning for chosing the current way)_
 
 ### This change affects (describe how if yes)
 - [ ] Performance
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,80 +1,35 @@
...
-#### What alternative approaches were there?
-_(briefly list any if applicable)_
-
-
-
-#### What feature flags were used to cover this change?
-_(briefly list any if applicable)_
+### What feature flags were used to cover this change?
 
+### What alternative approaches were there?
```
bmartel, line 16:
```suggestion
```

I think this information would be best kept in a TDD or another document. As we had discussed the heft of this template, it had been also mentioned this was duplicating that effort in the best case, worst case was just too much noise for the day to day work.

.github/pull_request_template.md
```diff
@@ -1,80 +1,35 @@
...
-- [ ] unit
-
-
+- [ ] e2e (codecept)
+- [ ] integration (cypress)
+- [ ] unit (jest)
 
 ### Which logical domain(s) does this change affect?
-_(for bug fixes/features, be as precise as possible. ex. Authentication, Annotation History, Review Stream etc.)_
+(comma-separated list like "Taxonomy, Dynamic Children")
```
bmartel, line 35: This one as well I think in the context of OSS, but also our own ability to use this information, would it not be better to just allow for the **Reason for change** section to facilitate this? As part of the reason for change, I often find myself writing in a lot of these other sections information, and in general I don't see a lot of people filling it in. I think if we collapse this all the way down, we can then use **Reason for change** as the litmus test of the reviewers ability to understand what they are reviewing. If the change is not obvious after seeing the excerpt describing the change, or it doesn't align well with the code, then it is a single point to call out to get alignment and clarity which helps cut down on the amount of QA reopens.

.github/pull_request_template.md
```diff
@@ -1,80 +1,35 @@
+Main description of what's done here, shortly, with reasoning, references, thoughts and ideas for better understanding.
+This text is intended to be removed and replaced with actual description. Other items can also be removed or edited if not relevant. If any of following items require more description, it should also go in this block.
+Please, read all items and make an action every time to create more conscious PR. Thanks!
```
bmartel, line 3:
```suggestion
Please, read all items and make an action every time to create more consistent pull requests. Thanks!
```

hlomzik: My idea is to remove non relevant sections, that's why I use "action" here, developer can edit, remove, fill in or just tick a checkbox. Does it make sense?

.github/pull_request_template.md
hlomzik: Huge plus on ability to use this info! This section is really no use for anything, thanks.
About the rest — yes, I'm also fan of good descriptions with all related info there, and I even tried to make such description placeholder, that it will push people towards more thoughts and nuances.

Gondragos: For me, it helps to systematize what has been done. And also shows whether the changes were intended. It can be especially useful if debugging an existing problem leads us to the specific PR.

In general, it is not about the cause, but about what has been affected. It's the same as tags for feature-flags. If tags are good you can realize which FF can cause the problem you have found.

hlomzik: About FFs — yes, because there is a search by tags. But how this list will help us and when will it be not deferred from name and description?

.github/pull_request_template.md
hlomzik: One of the main reasons to read old PRs is to understand what and why was done. And this section will help to understand why.

bmartel (approved):
Looks amazing. We should be using this as soon as possible 🔥

PR #1640, opened 2023-12-11
fix: LEAP-443: Select only one annotation on init

hlomzik:
### Main change

**Only one annotation is selected during task opening.**

We have the concept of selecting annotation, when we not just assign it as current one, but also update data in tags, trigger external event, load annotation history, setup hotkeys, set initial values.

Previosly it was required to select every annotation to do some extra work in regions/tags inside it because of some quirky legacy code. I fixed what I found so far and ran all possible tests — they all are green, so consider this change mostly safe.

Benefits:
- improved performance as only one annotation is selected, reducing number of calculations and renders
- no excess network requests to retrieve annotation history for every of these annotations
- more clean code
- one more step to remove one of the oldest legacy concept of `states`: it's not used in init anymore

Things changed:
- `STORE_INIT_OK` global var added to catch tags that can't properly work with new system (edge cases we missed)
- apparently we have only one unit test for tags — Ranker; it's securely mocked

Concerns:
- `states` in Video tag select only *Labels controls, but changed code is in `fixBrokenAnnotation` which will do nothing to Video anyway
- in potential missed edge case app will just crash
- we could potentially check that we are inside init process by some flag, right?

### Other changes
- removed Predictions panel and related code like showAllPredictions
- removed panels from App and down the stream — we don't use them anywhere (3rd party LSF can be affected, but not sure anyone using panels there)
- converted `LabelStudio.js` to TypeScript
- fixed bunch of annoying react warnings (non-html props, wrong values, missed refs)
- started to reorganize our "envs" to get rid of them at some point; for now it's `getRoot()` removed from there
- removed `SidebarPage` as it was useless, plus removed multi-tabs support from `SidebarTabs` as `panels` were not used; plus that's all only for the old interface which is about to be removed
- removed `hydrated` as it's not used anymore after https://github.com/HumanSignal/label-studio-frontend/pull/1166
- changed order of annotations to original one without reversing; thus order will always be the same, fixing the issue when list was reversed again after task reload; visual order is defined in `AnnotationsCarousel` anyway, so no changes in UX
- fixed annotations order in View All to be the same as in annotation tabs

- [ ] Tests for the changes have been added/updated

LSF initialization is slow so that's one of the first steps of big effort to make it lightweight and streamlined. Need for selecting every annotation even if it's not displayed was part of legacy way of working with labels and other store parts with no initial architecture.

### What feature flags were used to cover this change?
`fflag_fix_front_leap_443_select_annotation_once` covers changes AppStore init

- [x] Performance — one annotation select, no excess code runs and network requests
- [ ] Security
- [ ] UX — of of the main goals is that nothing should be changed from user perspective

- [x] Yes, and covered entirely by feature flag(s)
- [ ] No

- [ ] e2e (codecept)
- [ ] integration (cypress)
- [ ] unit (jest)


src/LabelStudio.tsx
```diff
@@ -1,55 +1,98 @@
...
-import { destroy } from 'mobx-state-tree';
 import { destroy as destroySharedStore } from './mixins/SharedChoiceStore/mixin';
-import { cleanDomAfterReact, findReactKey } from './utils/reactCleaner';
+import { EventInvoker } from './utils/events';
 import { FF_LSDV_4620_3_ML, isFF } from './utils/feature-flags';
+import { cleanDomAfterReact, findReactKey } from './utils/reactCleaner';
+import { isDefined } from './utils/utilities';
+
+declare global {
+  interface Window { Htx: any }
```
yyassi-heartex, line 19: do we want to continue exposing Htx? are we or any one actually using this? im wondering if we want to only expose this variable on a specific use case like while debugging. maybe we can look at only making Htx available if user has added debug=true to the url when loading the page. this'll help clean things up yet still allow us to use this when debugging. still entirely dependant on if Htx is used anywhere

yyassi-heartex: did a quick search, I do see that it is used with in editor, editor/e2e and refrenced in LSP ReviewPage. but now that we're moving to a mono-repo setup we maybe able to move away from this direction, open to discussion on this

src/components/HtxTextBox/HtxTextBox.js
```diff
@@ -150,8 +165,11 @@ export class HtxTextBox extends React.Component {
       isEditable,
       isDeleteable,
       text,
+
+      // don't pass non-DOM props to Paragraph
       ignoreShortcuts: _,
       onlyEdit: __,
+
```
yyassi-heartex, line 172: this is nit-picky, but its not really needed
```suggestion
```

src/components/Panel/Panel.js
```diff
@@ -27,7 +27,7 @@ export default observer(({ store }) => {
   const classname = [
     styles.block,
     styles.block__controls,
-    store.annotationStore.viewingAllAnnotations ? styles.hidden : '',
+    store.annotationStore.viewingAll ? styles.hidden : '',
   ].join(' ');
```
yyassi-heartex, line 31: we should probably filter out empty strings/falsey values from here - but again this is nit-picky

src/components/SidebarTabs/SidebarTabs.js
```diff
@@ -1,43 +1,14 @@
...
 
-const SidebarContext = createContext();
-
-export const SidebarTabs = observer(({ active, children }) => {
-  const [selected, setSelected] = useState(active);
-  const tabs = Children.toArray(children);
-
+// @todo there was an idea of switchable tabs, but they were not used,
+// @todo so implementation was removed and the whole part of interface
+// @todo is waiting to be removed in favor of new UI (see FF_DEV_3873)
```
yyassi-heartex, line 7: I see this FF is on for all but 5 users on production - what do we need to do to get buy in for the remaining 5 so we can archive this FF and remove the outdated code? this'll help clean up this portion of the code

src/LabelStudio.tsx
hlomzik: Thank you, that's important question! **tl;dr**: we need it
1. It's pretty common for libraries to expose their global(s) to access features from anywhere. There are different approaches of course, but that's not unusual and bad practice.
2. It really helped lot of times to deal and investigate issues — at any point we can get access to virtually any value from the console. We can ask anyone to do this at any time, especially in the middle of labeling without any reloading.
3. Unfortunately we use `Htx` in some fundamential places, the main one is detection of region type in [MST model Area](https://github.com/HumanSignal/label-studio-frontend/blob/f209200de05aad01cba5dffc59883d0de4615458/src/regions/Area.js#L55), plus couple of other places.
4. If we leave it anyway, but cover with some condition, this will leave the same code plus a little more. And I don't think this var is conflicting with anything. Tell me if you have objections and second thoughts!

src/components/Panel/Panel.js
hlomzik:
```suggestion
  ].filter(Boolean).join(' ');
```

Gondragos (approved):

Gondragos (approved):

src/components/App/App.js
```diff
@@ -165,8 +165,14 @@ class App extends Component {
 
   renderAllAnnotations() {
     const as = this.props.store.annotationStore;
+    const entities = [...as.annotations, ...as.predictions];
 
-    return <Grid store={as} annotations={[...as.annotations, ...as.predictions]} root={as.root} />;
+    if (isFF(FF_SIMPLE_INIT)) {
+      // the same sorting we have in AnnotationsCarousel, so we'll see the same order in both places
+      entities.sort((a, b) => new Date(b.createdDate).getTime() - new Date(a.createdDate).getTime());
```
yyassi-heartex, line 172: if we have the same sorting function being used, maybe a good idea to define it some where in `lsf/src/utils/date.js` or `utilities.js` and reuse it in both spots from there - chance are if the code needs to change down the line we'll need to update both locations at the same time

PR #1649, opened 2023-12-21
fix: LEAP-447: Fix Number tag validation 

Gondragos:

_(check all that apply)_
- [ ] Product design
- [x] Frontend

The change was necessary due to failures of the number validation in some cases that resulted in the ability to save invalid values.

#### What does this fix?
This fix introduces value validation for the Number tag, conducts a refactoring of the existing validation code, and implements a per-item mode validation.

Nope

Nope

N/A

N/A

_(check only one)_
- [x] No
At least it should not.

_(check all that apply)_
- [ ] e2e
- [x] integration
- [ ] unit

`Number`, `DateTime`, `Validation`, `Required`, `perItem`

### DIFF
--- a/src/mixins/PerItem.js
+++ b/src/mixins/PerItem.js
@@ -1,4 +1,5 @@
 import { types } from 'mobx-state-tree';
+import { FF_LSDV_4583, isFF } from '../utils/feature-flags';
 
 /**
  * This mixing defines perItem control-tag's parameter and related basic functionality
@@ -22,6 +23,26 @@ const PerItemMixin = types
     },
   }))
   .actions(self => ({
+    _validatePerItem() {
+      const objectTag = self.toNameTag;
+
+      return self.annotation.regions
+        .every((reg) => {
+          const result = reg.results.find(s => s.from_name === self);
+
+          if (!result || !result.hasValue) {
+            return true;
+          }
+          const value = result.mainValue;
+          const isValid = self.validateValue(value);
+
+          if (!isValid) {
+            objectTag.setCurrentItem(reg.item_index);
+            return false;
+          }
+          return true;
+        });
+    },
     createPerItemResult() {
       self.createPerObjectResult({
         item_index: self.toNameTag.currentItemIndex,
--- a/src/mixins/PerItem.js
+++ b/src/mixins/PerItem.js
@@ -22,6 +22,15 @@ const PerItemMixin = types
     },
   }))
   .actions(self => ({
+    /**
+     * Validates all values related to the current classification per item (Multi Items Segmentation).
+     *
+     * - This method should not be overridden.
+     * - It is used only in validate method of the ClassificationBase mixin.
+     *
+     * @returns {boolean} - The validated value.
+     * @private
+     */
     _validatePerItem() {
       const objectTag = self.toNameTag;
 
--- a/src/mixins/PerItem.js
+++ b/src/mixins/PerItem.js
@@ -1,5 +1,4 @@
 import { types } from 'mobx-state-tree';
-import { FF_LSDV_4583, isFF } from '../utils/feature-flags';
 
 /**
  * This mixing defines perItem control-tag's parameter and related basic functionality
--- a/src/mixins/PerItem.js
+++ b/src/mixins/PerItem.js
@@ -28,7 +28,7 @@ const PerItemMixin = types
      * - This method should not be overridden.
      * - It is used only in validate method of the ClassificationBase mixin.
      *
-     * @returns {boolean} - The validated value.
+     * @returns {boolean}
      * @private
      */
     _validatePerItem() {
--- a/src/mixins/Required.js
+++ b/src/mixins/Required.js
@@ -6,71 +6,75 @@ const RequiredMixin = types
     required: types.optional(types.boolean, false),
     requiredmessage: types.maybeNull(types.string),
   })
-  .actions(self => ({
-    validate() {
-      if (!self.required) return true;
+  .actions(self => {
+    const Super = {
+      validate: self.validate,
+    };
 
-      if (self.perregion) {
+    return {
+      validate() {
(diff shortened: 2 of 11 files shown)


src/mixins/PerItem.js
```diff
@@ -22,6 +22,26 @@ const PerItemMixin = types
...
   }))
   .actions(self => ({
+    _validatePerItem() {
+      const objectTag = self.toNameTag;
+
+      return self.annotation.regions
+        .every((reg) => {
+          const result = reg.results.find(s => s.from_name === self);
+
+          if (!result || !result.hasValue) {
```
yyassi-heartex, line 32: this should give us the same outcome
```suggestion
          if (!result?.hasValue) {
```

src/mixins/PerItem.js
```diff
@@ -22,6 +22,26 @@ const PerItemMixin = types
     },
   }))
   .actions(self => ({
+    _validatePerItem() {
```
hlomzik, line 25: what's the idea behind underscore here? we use underscores rarely and for purely internal things, and those are called from other models using this mixin.

src/tags/control/ClassificationBase.js
```diff
@@ -40,6 +41,23 @@ const ClassificationBase = types.model('ClassificationBase', {
...
   }).actions(self => {
     return {
+      validate() {
+        if (self.perregion) {
+          return self._validatePerRegion();
+        }
+        else if (self.peritem && isFF(FF_LSDV_4583)) {
+          return self._validatePerItem();
+        }
+        else {
```
hlomzik, line 51: that's something new! we don't use such style for if-else and I thought we have this covered by linting. if not, we have to add it to the list of linting improvements

src/tags/control/Number.js
```diff
@@ -72,80 +72,113 @@ const Model = types
...
+      },
 
-    onHotKey() {
-      return self.increaseValue();
-    },
-  }));
+      onHotKey() {
+        return self.increaseValue();
+      },
+    }; });
```
hlomzik, line 181: another linting issue

src/tags/control/ClassificationBase.js
```diff
@@ -40,6 +41,23 @@ const ClassificationBase = types.model('ClassificationBase', {
...
+          return self._validatePerObject();
+        }
+      },
+      validateValue() {
```
hlomzik, line 55: comment on underscores: feels especially wrong to see those methods one by one with different styles

hlomzik (changes requested):
leaving only linting comments for now, but blocking PR to check the actual validation, change here is pretty big and important

src/tags/control/Number.js
```diff
@@ -72,80 +72,113 @@ const Model = types
...
+        const errors = [];
 
-      // let's fix only required perRegions
-      if (self.perregion && self.required) {
-        const object = self.toNameTag;
+        if (isDefined(self.min) && value < self.min) {
+          errors.push(`Value must be greater that or equal to ${self.min}`);
+        }
+        if (isDefined(self.max) && value > self.max) {
+          errors.push(`Value must be less that or equal to ${self.max}`);
```
hlomzik, line 89: > greater and less tha**n**

src/tags/control/Number.js
```diff
@@ -72,80 +72,113 @@ const Model = types
...
+    const Super = { validateValue: self.validateValue };
 
-    needsUpdate() {
-      if (self.result) self.number = self.result.mainValue;
-      else self.number = null;
-    },
+    return {
+      validateValue(value) {
+        if (!Super.validateValue(value)) return false;
+        if (!isDefined(value)) return true;
```
hlomzik, line 81: should not those two lines be swapped?

src/tags/control/Number.js
```diff
@@ -72,80 +72,113 @@ const Model = types
...
-    onChange(e) {
-      const value = +e.target.value;
+      onChange(e) {
+        const value = +e.target.value;
 
-      if (!isNaN(value)) self.setNumber(value);
-    },
+        if (!isNaN(value)) {
+          self.setNumber(value);
+          e.target.value = isDefined(self.number) ? self.number : '';
```
hlomzik, line 154: why do we need this?

src/mixins/PerItem.js
Gondragos: The idea was to prevent using this method outside of the bundle `PerItemMixin <-> ClassificationBase`. I'm not sure how to prevent it in better way.

src/tags/control/ClassificationBase.js
Gondragos: It's great 'cause `validateValue` should be overridden and `_validatePerObject` should not. It's good if you can feel that something here must to be taken in account.

src/tags/control/Number.js
Gondragos: I think that it's better to put `Super`'s  method call at the beginning. It may be called in chain with some other mixing and that can become complicated.

src/tags/control/Number.js
Gondragos: It should prevent having values with leading zeros. That was annoing.

src/tags/control/ClassificationBase.js
hlomzik: add a comment that this method is the one to override in inheritors, not `validate()`

hlomzik (approved):

yyassi-heartex (approved):

src/mixins/Required.js
```diff
@@ -6,71 +6,76 @@ const RequiredMixin = types
...
+          }
         }
-      }
-
-      return true;
-    },
-  }));
+        return true;
+      },
+    }; });
```
hlomzik, line 79:
```suggestion
    };
  });
```

hlomzik (approved):

PR #1653, opened 2023-12-29
fix: LEAP-248: Fix app crashing by hotkeys 

Gondragos:
This PR 
1. wraps unbinding functional to separate keys corectly outside of the keymaster
2. adds alias for `,` to avoid problems with it
3. improves tokenization of hotkeys

- [x] Tests for the changes have been added/updated

Keymaster can't handle unbinding of multiple keys. It also misunderstands `,` and thinks that it is two hotkeys in one (`','` and `''`). So that `,` breaks application in two ways related to each other.

### What alternative approaches were there?
1. we can fix keymaster in a fork (but we did it before and we didn't like it)
2. we can use another library to handle hotkeys. But it's hard to test and integrate.

- [ ] Performance
- [ ] Security
- [ ] UX

- [x] No

- [ ] e2e (codecept)
- [x] integration (cypress)
- [ ] unit (jest)

`Hotkeys`, `Labels`

### DIFF
--- a/src/core/Hotkey.ts
+++ b/src/core/Hotkey.ts
@@ -82,6 +82,7 @@ keymaster.filter = function(event) {
 const ALIASES = {
   'plus': '=', // "ctrl plus" is actually a "ctrl =" because shift is not used
   'minus': '-',
+  ',': '¼', // This should be a comma but it's not working with keymaster
 };
 
 export const Hotkey = (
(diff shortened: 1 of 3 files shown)


src/core/Hotkey.ts
```diff
@@ -82,6 +82,7 @@ keymaster.filter = function(event) {
 const ALIASES = {
   'plus': '=', // "ctrl plus" is actually a "ctrl =" because shift is not used
   'minus': '-',
+  ',': '¼', // This should be a comma but it's not working with keymaster
```
hlomzik, line 85: why is it `¼`? pretty confusing

hlomzik (approved):

PR #1665, opened 2024-01-11
fix: OPTIC-353: Fix null errors when switching from task detail to settings

Travis1282:
[OPTIC-353](https://humansignal.atlassian.net/browse/OPTIC-353)

Error when switching from task detail to settings from a video annotation with the region or intermittently images with regions fixed with a null block of child nodes before they are used as an arg in clearRenderedApp. Note this bug is connected to the effort to reduce the number of memory leaks and can be also resolved by turning off the feature flag: fflag_fix_front_lsdv_4620_memory_leaks_100723_short 

[OPTIC-353]: https://humansignal.atlassian.net/browse/OPTIC-353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/LabelStudio.js
+++ b/src/LabelStudio.js
@@ -79,10 +79,10 @@ export class LabelStudio {
     };
 
     const clearRenderedApp = () => {
+      if (!rootElement.childNodes?.length) return;
+
       const childNodes = [...rootElement.childNodes];
       // cleanDomAfterReact needs this key to be sure that cleaning affects only current react subtree
-
-      if (!childNodes[0]) return;
       const reactKey = findReactKey(childNodes[0]);
 
       unmountComponentAtNode(rootElement);
--- a/src/LabelStudio.js
+++ b/src/LabelStudio.js
@@ -81,6 +81,8 @@ export class LabelStudio {
     const clearRenderedApp = () => {
       const childNodes = [...rootElement.childNodes];
       // cleanDomAfterReact needs this key to be sure that cleaning affects only current react subtree
+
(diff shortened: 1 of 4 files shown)


src/LabelStudio.js
```diff
@@ -81,6 +81,8 @@ export class LabelStudio {
     const clearRenderedApp = () => {
       const childNodes = [...rootElement.childNodes];
       // cleanDomAfterReact needs this key to be sure that cleaning affects only current react subtree
+
+      if (!childNodes[0]) return;
```
hlomzik, line 85:
```suggestion
      if (!rootElement.childNodes?.length) return;

      const childNodes = [...rootElement.childNodes];
      // cleanDomAfterReact needs this key to be sure that cleaning affects only current react subtree
```

hlomzik: to not mess with the comment and exit earlier

src/stores/AppStore.js
```diff
@@ -880,7 +880,7 @@ export default types
 
         let node;
 
-        while ((node = children.shift())) {
+        while ((node = children.shift() && children)) {
```
hlomzik, line 883: if `children` is undefined it will fail on `shift()`
and if it's some magic which works for some reasons, it's better to document it!

bmartel (approved):

hlomzik (approved):

PR #1689, opened 2024-02-02
fix: LEAP-640: Don't save empty draft on View All

hlomzik:
We should save draft when user clicks on View All, but we should save it only if there are changes made to annotation.

- [ ] Tests for the changes have been added/updated

Draft is created/updated **every** time user click on View All, even if that's fresh new task, if there are no changes, if draft was saved a second ago.

### What alternative approaches were there?
We could add checks for actual changes to `saveDraftImmediatelyWithResults()`, but the whole approach should be revisited. So current fix is a quick patch for the actual problem.
Also the main reason for draft not being saved is the check for `editable` in `saveDraft()` — at this moment View All is already displayed and all annotations are not editable. Most probably it will be enough just to rewrite this check to check for annotation explicitly.

- [ ] Performance
- [ ] Security
- [x] UX — toast about "Draft saved" disappeared but it was only introduced a month ago; we'll return it later

- [x] No

- [ ] e2e (codecept)
- [ ] integration (cypress)
- [ ] unit (jest)

### DIFF
--- a/src/components/TopBar/TopBar.js
+++ b/src/components/TopBar/TopBar.js
@@ -1,15 +1,17 @@
 import { observer } from 'mobx-react';
+
+import { IconViewAll, LsPlus } from '../../assets/icons';
+import { Button } from '../../common/Button/Button';
+import { Tooltip } from '../../common/Tooltip/Tooltip';
 import { Block, Elem } from '../../utils/bem';
+import { FF_DEV_3873, isFF } from '../../utils/feature-flags';
+import { AnnotationsCarousel } from '../AnnotationsCarousel/AnnotationsCarousel';
 import { DynamicPreannotationsToggle } from '../AnnotationTab/DynamicPreannotationsToggle';
 import { Actions } from './Actions';
 import { Annotations } from './Annotations';
 import { Controls } from './Controls';
 import { CurrentTask } from './CurrentTask';
-import { FF_DEV_3873, isFF } from '../../utils/feature-flags';
-import { Button } from '../../common/Button/Button';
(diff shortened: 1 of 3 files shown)


src/stores/Annotation/store.js
```diff
@@ -50,6 +50,19 @@ const AnnotationStoreModel = types
     function toggleViewingAll() {
       if (self.viewingAllAnnotations || self.viewingAllPredictions) {
         if (self.selected) {
+          // const comments = self.store.commentStore;
+
+          // @todo current comment is an object and that was not a part of original fix
+          // @todo so leave it for later
```
hlomzik, line 56:
```suggestion
          // @todo `currentComment` is an object and saving them was not a part of original fix
          // @todo so I leave it for next fix coming soon
```

src/components/TopBar/TopBar.js
```diff
@@ -1,15 +1,17 @@
...
 import { Actions } from './Actions';
 import { Annotations } from './Annotations';
 import { Controls } from './Controls';
 import { CurrentTask } from './CurrentTask';
-import { FF_DEV_3873, isFF } from '../../utils/feature-flags';
-import { Button } from '../../common/Button/Button';
-import { Tooltip } from '../../common/Tooltip/Tooltip';
-import { IconViewAll, LsPlus } from '../../assets/icons';
-import { AnnotationsCarousel } from '../AnnotationsCarousel/AnnotationsCarousel';
+
```
yyassi-heartex, line 14:
```suggestion
```

yyassi-heartex (approved):

yyassi-heartex (approved):

